PR #6, opened 2019-12-16
Replace Byowl by fio-distributed

rsevilla87:
Replace Byowl by fio-distributed  so the benchmark would be able to send metrics to the ES endpoint


workloads/run_etcd_tests_fromgit.sh
```diff
@@ -181,28 +181,41 @@ EOF
...
+    port: ${_es_port}
+  fio_path: /var/tmp
+  clustername: ${cloud_name}
   test_user: ${cloud_name}-ci
   metadata_collection: true
   metadata_sa: backpack-view
   metadata_privileged: true
   workload:
-    name: byowl
+    name: fio_distributed
```
jtaleric: I previously had the fio_distributed work here, however it did not work :/

I saw if we bump our FIO version it might be fixed?

PR #10, opened 2020-01-23
add multiple client-server pair test case

mohit-sheth:
This will give us the ability to run uperf test with 1,2 and 4 client-server pairs from a single script
@jtaleric


mohit-sheth:
@jtaleric do not merge, will add a loop for the pairs

PR #12, opened 2020-02-07
add metadata_collection as ENV var and increase the timeout

mohit-sheth:
@jtaleric


jtaleric:
We still have conflicts

PR #13, opened 2020-02-11
Accept tolerance as an arg

mohit-sheth:
also added a the ability to compare both 'stream' and 'rr' numbers against a reference. @jtaleric

### DIFF
--- a/workloads/network-perf/compare.py
+++ b/workloads/network-perf/compare.py
@@ -25,6 +25,9 @@ parser.add_argument("--tests", help="Test to compare",nargs=1,choices=['stream',
                     dest="tests")
 parser.add_argument("--protocol", help="Protocol to compare",nargs=1,choices=['tcp','udp'],required=True,
                     dest="protocol")
+parser.add_argument("--tolerance", help="Accepted deviation (+/-) from the reference result",nargs=1,required=False,
+                    dest="tolerance", default="5")
+
 params = parser.parse_args()
 data = None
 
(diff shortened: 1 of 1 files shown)


workloads/network-perf/compare.py
```diff
@@ -25,6 +25,9 @@
                     dest="tests")
 parser.add_argument("--protocol", help="Protocol to compare",nargs=1,choices=['tcp','udp'],required=True,
                     dest="protocol")
+parser.add_argument("--tolerance", help="Exit code will be 1/0 based on the number of regressions",nargs=1,required=True,
```
jtaleric: required should be `False`, lets default=5

workloads/network-perf/compare.py
```diff
@@ -40,14 +43,38 @@
...
     exit(1)
 
+def percent_change(value, reference):
+    if reference:
+        return abs(((value-reference)*1.0/reference))*100
+    else:
+        return -1
+
+def compare(data, tolerance, latency=False):
+    ref_val = data[params.uuid[0]]
```
jtaleric: does `params.uuid` have scope here? Shouldn't we be passing this as a param to `compare`?

workloads/network-perf/compare.py
```diff
@@ -56,8 +83,12 @@ def compare(data):
...
-                comparison = compare(result[main_metric[test_type]])
+                if test_type == "rr" :
+                    comparison = compare(result[main_metric[test_type]], int(params.tolerance[0]), latency=True)
+                else:
+                    comparison = compare(result[main_metric[test_type]], tolerance=int(params.tolerance[0]))
                 if params.uuid[0] not in comparison :
-                    print("Test Failure : UPerf reports lower throughput for, Test: {}, Protocol {}, Message_size {}, Threads {}".format(
+                    print("Test Failure : UPerf reports regression for, Test: {}, Protocol {}, Message_size {}, Threads {}".format(
                         test_type,proto,size,thread))
                     exit(1)
```
jtaleric: I suggest, lets capture the differences (for all message sizes and threads) and present them at the end. Versus exiting right after finding one difference. We should capture any possible regression.

workloads/network-perf/compare.py
```diff
@@ -56,8 +84,14 @@ def compare(data):
...
                 result = data['test_type.keyword'][test_type]['protocol'][
                         proto]['message_size'][size]['num_threads'][thread]
-                comparison = compare(result[main_metric[test_type]])
+                if test_type == "rr" :
+                    comparison = compare(result[main_metric[test_type]], int(params.tolerance[0]), latency=True)
+                else:
+                    comparison = compare(result[main_metric[test_type]], tolerance=int(params.tolerance[0]))
                 if params.uuid[0] not in comparison :
-                    print("Test Failure : UPerf reports lower throughput for, Test: {}, Protocol {}, Message_size {}, Threads {}".format(
+                    print("\nTEST FAILURE: Uperf reports regrerssion which is beyond the acceptable deviation of {}%".format(params.tolerance[0]))
```
jtaleric: I think you can safely remove the `\n` from the output.

PR #16, opened 2020-02-13
Run compare across two UUIDs

jtaleric:
Compare the results of two UPerf UUIDs

### DIFF
--- a/workloads/network-perf/run_network_compare.sh
+++ b/workloads/network-perf/run_network_compare.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+datasource="elasticsearch"
+tool="uperf"
+function="compare"
+if [ "$#" -ne 2 ]; then
+  echo "Syntax error : Script expects two UUIDs"
+  echo "               ./run_network_compare.sh <baseline_uuid> <new_uuid>"
+  exit 1
+fi
+base_uuid=$1
+compare_uuid=$2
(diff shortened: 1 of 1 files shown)


workloads/network-perf/run_network_compare.sh
```diff
@@ -0,0 +1,51 @@
...
+if [ "$es" == "" ]; then
+  echo "Unable to execute compare - no ES passed"
+  exit 1
+fi
+
+git clone https://github.com/cloud-bulldozer/touchstone
+cd touchstone
+python -m venv ./compare
+source ./compare/bin/activate
+pip install pyyaml
```
aakarshg: can this be updated to `pip install -r requirements.txt` ?

jtaleric: yup!

workloads/network-perf/run_network_compare.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+set -x
+
+datasource="elasticsearch"
+tool="uperf"
+function="compare"
+base_uuid=$1
```
dry923: Minor nit but if you know the total number of arguments to expect why not just check for that instead of multiple if's?

jtaleric: Smart! Too many after thoughts, thanks!

PR #18, opened 2020-02-24
add the ability to compare 2 uuids from the scripts itself

mohit-sheth:
@jtaleric @chaitanyaenr
This commit gives us the ability to compare the test_uuid with the baseline_uuid if we set COMPARE=true. Also this commit adds a tuning to specify throughput and latency tolerances individually (default=5). This is because latency can vary a lot depending on the environment (ex. 4.3 perfscale tests). 
The baseline_uuid vars for different tests will be set once they are finalised

### DIFF
--- a/workloads/network-perf/README.md
+++ b/workloads/network-perf/README.md
@@ -1,0 +1,103 @@
+# Network Scripts
+
+The purpose of the network scripts is to run uperf workload on the Openshift Cluster.
+There are 4 types network tests:
+1. pod to pod using Hostnetwork
+2. pod to pod using SDN
+3. pod to pod using Service
+4. pod to pod using Multus (NetworkAttachmentDefinition needs to be provided)
+
+Running from CLI:
+
+```sh
+$ ./run_<test-name>_network_test_fromgit.sh 
+```
+
+## Environment variables
+
+### ES_SERVER
+Default: `search-cloud-perf-lqrf3jjtaqo7727m7ynd2xyt4y.us-west-2.es.amazonaws.com`  
+Public elasticsearch server
+
+### ES_PORT
+Default: `80`  
+Port number for public elasticsearch server
+
+### METADATA_COLLECTION
+Default: `true`   
+Enable/Disable collection of metadata
+
+### COMPARE
+Default: `false`   
+Enable/Disable the ability to compare two uperf runs. If set to `true`, the next set of environment variables pertaining to the type of test are required
+
+### ES_SERVER_BASELINE 
+Default: `search-cloud-perf-lqrf3jjtaqo7727m7ynd2xyt4y.us-west-2.es.amazonaws.com`   
+Elasticsearch server used used by the baseline run 
+
+### ES_PORT_BASELINE
+Default: `80`  
+Port number for the elasticsearch server used by the baseline run
+
+### BASELINE_HOSTNET_UUID
+Default: ``   
+Baseline UUID for hostnetwork test  
+
+### BASELINE_POD_1P_UUID
+Default: ``   
+Baseline UUID for pod to pod using SDN test with 1 uperf client-server pair
+
+### BASELINE_POD_2P_UUID
+Default: ``   
+Baseline UUID for pod to pod using SDN test with 2 uperf client-server pair
+
+### BASELINE_POD_4P_UUID
+Default: ``   
+Baseline UUID for pod to pod using SDN test with 4 uperf client-server pair
+
+### BASELINE_SVC_1P_UUID
+Default: ``   
+Baseline UUID for pod to pod using service test with 1 uperf client-server pair
+
+### BASELINE_SVC_2P_UUID
+Default: ``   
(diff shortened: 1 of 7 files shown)


workloads/network-perf/run_multus_network_tests_fromgit.sh
```diff
@@ -135,6 +140,17 @@ if [ "$uperf_state" == "1" ] ; then
   exit 1
 fi
 
+if [[ ${COMPARE} == "true" ]] ; then
+  baseline_uperf_uuid=${_baseline_multus_uuid}
+  if [ "$_metadata_collection" == true ] ; then
```
aakarshg: We should do a json query to extract uuid, it's simpler and the code would be much more stable.

workloads/network-perf/run_pod_network_test_fromgit.sh
```diff
@@ -114,6 +129,23 @@ if [ "$uperf_state" == "1" ] ; then
...
 
+if [[ ${COMPARE} == "true" ]] ; then
+  if [ "${pairs}" == "1" ] ; then
+    baseline_uperf_uuid=${_baseline_pod_1p_uuid}
+  elif [ "${pairs}" == "2" ] ; then
+    baseline_uperf_uuid=${_baseline_pod_2p_uuid}
+  elif [ "${pairs}" == "4" ] ; then
+    baseline_uperf_uuid=${_baseline_pod_4p_uuid}
+  fi
+  if [ "$_metadata_collection" == true ] ; then
```
aakarshg: would be best to use json query

workloads/network-perf/run_serviceip_network_test_fromgit.sh
```diff
@@ -114,6 +129,23 @@ if [ "$uperf_state" == "1" ] ; then
...
 
+if [[ ${COMPARE} == "true" ]] ; then
+  if [ "${pairs}" == "1" ] ; then
+    baseline_uperf_uuid=${_baseline_svc_1p_uuid}
+  elif [ "${pairs}" == "2" ] ; then
+    baseline_uperf_uuid=${_baseline_svc_2p_uuid}
+  elif [ "${pairs}" == "4" ] ; then
+    baseline_uperf_uuid=${_baseline_svc_4p_uuid}
+  fi
+  if [ "$_metadata_collection" == true ] ; then
```
aakarshg: json query here too

workloads/network-perf/run_multus_network_tests_fromgit.sh
mohit-sheth: agreed

workloads/network-perf/run_network_compare.sh
```diff
@@ -9,38 +9,38 @@ if [ "$#" -ne 2 ]; then
...
-pip install -r requirements.txt
+pip3 install -r requirements.txt
 python3 setup.py develop
 if [[ $? -ne 0 ]] ; then
   echo "Unable to execute compare - Failed to install touchstone"
   exit 1
 fi
 
-touchstone_compare $tool $datasource ripsaw -url $es -u $base_uuid $compare_uuid -o yaml | tee ../compare.yaml
+touchstone_compare $tool $datasource ripsaw -url $es_server:$es_port -u $base_uuid $compare_uuid -o yaml | tee ../compare.yaml
```
jtaleric: for an artifact , could we generate the csv (I need to get that added to touchstone, I already have a PR to send..)

workloads/network-perf/run_network_compare.sh
```diff
@@ -2,45 +2,65 @@
...
-python compare.py --result compare.yaml --uuid $base_uuid --test stream --protocol udp || failed=1
+python3 compare.py --result compare.yaml --uuid $base_uuid --test stream --protocol udp --tolerance $throughput_tolerance || failed=1
 echo "Checking RR TCP"
-python compare.py --result compare.yaml --uuid $base_uuid --test rr --protocol tcp || failed=1
+python3 compare.py --result compare.yaml --uuid $base_uuid --test rr --protocol tcp --tolerance $latency_tolerance || failed=1
 echo "Checking RR UDP"
-python compare.py --result compare.yaml --uuid $base_uuid --test rr --protocol tcp || failed=1
+python3 compare.py --result compare.yaml --uuid $base_uuid --test rr --protocol tcp --tolerance $latency_tolerance || failed=1
 echo "Compare complete"
 exit $failed
```
mohit-sheth: @jtaleric  a few small changes here, basically adding ES_SERVER_BASELINE & ES_PORT_BASELINE (default set to public instance). This is useful when the baseline and compare tests are on different ES server. Also adding a way to pass tolerance(default set to 5)

workloads/network-perf/README.md
```diff
@@ -0,0 +1,103 @@
+# Network Scripts
```
mohit-sheth: @chaitanyaenr @jtaleric

chaitanyaenr: Nice work with the docs @mohit-sheth.

PR #23, opened 2020-03-27
adds ability to check if the cluster is up before running the test

mohit-sheth:
Pipeline will report a failure(Azure) much sooner with this commit
@jtaleric 
@chaitanyaenr

### DIFF
--- a/workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
+++ b/workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
@@ -31,6 +31,14 @@ if [ "$cloud_name" == "" ]; then
   cloud_name="test_cloud"
 fi
 
+# check if cluster is up
+date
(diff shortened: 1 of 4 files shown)


workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
```diff
@@ -31,6 +31,14 @@ if [ "$cloud_name" == "" ]; then
   cloud_name="test_cloud"
 fi
 
+# check if cluster is up
+date
+oc get clusterversion
+if [ $? ! == "0" ]; then
```
chaitanyaenr: It needs to be [[ $? != 0 ]].

mohit-sheth: changed it to [ $? -ne 0]

PR #27, opened 2020-04-20
Update etcd workload to use snafu's FIO

rsevilla87:
Replace BYOWL by fio-distributed and use hostPath to improve etcd IO emulation.

### DIFF
--- a/workloads/etcd-perf/run_etcd_tests_fromgit.sh
+++ b/workloads/etcd-perf/run_etcd_tests_fromgit.sh
@@ -4,8 +4,9 @@ set -x
 trap "rm -rf /tmp/ripsaw" EXIT
 _es=search-cloud-perf-lqrf3jjtaqo7727m7ynd2xyt4y.us-west-2.es.amazonaws.com
 _es_port=80
-samples=5
-sample=0
+latency_th=${LATENCY_TH:-10000000}
+index=ripsaw-fio-results
+curl_body='{"_source": false, "aggs": {"max-fsync-lat-99th": {"max": {"field": "fio.sync.lat_ns.percentile.99.000000"}}}}'
 
 if [[ "${ES_SERVER}" ]]; then
   _es=${ES_SERVER}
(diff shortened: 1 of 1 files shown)


jtaleric:
Nice @rsevilla87 -- any thoughts on how we can keep the pass/fail criteria after the run?

jtaleric:
It isn't really pass / fail, but it gives the CI guys a easy way to see the latency reported : https://github.com/cloud-bulldozer/plow/pull/27/files#diff-5d34d0be0f9fc260c39c4cea6ee46d03L84-L90

rsevilla87:
@jtaleric I've added a threshold variable to be compared with the max 99th latency indexed in ES. Is that you you mean?

workloads/etcd-perf/run_etcd_tests_fromgit.sh
```diff
@@ -42,35 +43,44 @@ cat << EOF | oc create -n my-ripsaw -f -
...
-    server: $_es
-    port: $_es_port
-  clustername: $cloud_name
+    server: ${_es}
+    port: ${_es_port}
+  clustername: ${cloud_name}
   test_user: ${cloud_name}-ci
   metadata_collection: true
   metadata_sa: backpack-view
   metadata_privileged: true
```
dry923, line 56: Metadata should now be in the form:
metadata:
  collection: true
  serviceaccount: backpack-view
  privileged: true

etc

workloads/etcd-perf/run_etcd_tests_fromgit.sh
```diff
@@ -42,35 +43,44 @@ cat << EOF | oc create -n my-ripsaw -f -
...
+      samples: 5
+      servers: 1
+      jobs:
+        - write
+      bs:
+        - 2300
+      numjobs:
+        - 1
+      filesize: 22Mib
+  global_overrides:
```
dry923, line 72: I assume this is supposed to be aligned under workload?

workloads/etcd-perf/run_etcd_tests_fromgit.sh
rsevilla87: fixed!

workloads/etcd-perf/run_etcd_tests_fromgit.sh
rsevilla87: fixed!

PR #28, opened 2020-04-20
Generating CSV for uperf results and then pushing it to google sheets

mohit-sheth:
The google sheets part is optional, will only push if the user provides google service account key and an email to share the results with.
If we set COMPARE to true it will compare against another set of uuids and will also give us Percent Change and a Pass/Fail based on the tolerance provided by the user. The tolerance can be set through env vars and can be different for throughput and latency. 
In the example below the tolerance is set to 1% for both
https://docs.google.com/spreadsheets/d/1KVaU_S6jW5fmeLiOOJeZ0AmaNCD80vqvx43xRMTGzrI/edit?usp=sharing


workloads/network-perf/csv_gen.py
```diff
@@ -0,0 +1,194 @@
...
+    scope = [
+        "https://spreadsheets.google.com/feeds",
+        "https://www.googleapis.com/auth/drive",
+    ]
+    credentials = ServiceAccountCredentials.from_json_keyfile_name(google_svc_acc_key, scope)
+    gc = gspread.authorize(credentials)
+
+    sh = gc.create(params.sheetname)  # Specify name of the Spreadsheet
+    sh.share(email_id, perm_type="user", role="writer")
+    # sh.share('jtaleric@redhat.com', perm_type='user', role='writer')
```
smalleni, line 144: Do we want to expose Joe's email address ? ;)

workloads/network-perf/common.sh
```diff
@@ -3,7 +3,12 @@ set -x
 
 _es=search-cloud-perf-lqrf3jjtaqo7727m7ynd2xyt4y.us-west-2.es.amazonaws.com
 _es_port=80
+_es_baseline=search-cloud-perf-lqrf3jjtaqo7727m7ynd2xyt4y.us-west-2.es.amazonaws.com
+_es_baseline_port=80
+throughput_tolerance=5
+latency_tolerance=5
 _metadata_collection=true
+client_server_pairs=(1 2 4)
```
jtaleric, line 11: Can we default to 1? And in our CI if we need something different we just update this?

jtaleric: Ah, I see later there is a dependence on this..

mohit-sheth: the baremetal CI uses 1,2 and 4 pairs as default I guess
we can change this manually as our requirement

PR #33, opened 2020-05-01
Fixing metadata variable from sa to serviceaccount

dry923:
The service account variable was changed from sa to serviceaccount. This fixes it here.


mohit-sheth:
tested and it works!
thanks @dry923

PR #35, opened 2020-05-29
Add Cerberus check while running Uperf

mohit-sheth:
This commit will give us the ability to check the health of the cluster before starting uperf and also while it is running. If the cluster is unhealthy the workload will fail and exit out
Fixes https://github.com/cloud-bulldozer/plow/issues/34

### DIFF
--- a/workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
+++ b/workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
@@ -31,6 +31,7 @@ spec:
     collection: ${_metadata_collection}
     serviceaccount: backpack-view
     privileged: true
+  cerberus_url: "$CERBERUS_URL" 
   workload:
     name: uperf
     args:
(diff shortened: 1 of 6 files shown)


workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
```diff
@@ -62,6 +63,10 @@ sleep 30
 
 uperf_state=1
 for i in {1..240}; do
+  if [ "$(oc get benchmarks.ripsaw.cloudbulldozer.io -n my-ripsaw -o json | jq -r .items[].status.state)" == "Error" ]; then
+    echo "Cerberus status is False, Cluster is unhealthy"
+    break
```
chaitanyaenr, line 68: Maybe exit 1 instead of break?

chaitanyaenr:
@dry923 PTAL.

workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
mohit-sheth: done

workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
```diff
@@ -62,6 +63,10 @@ sleep 30
 
 uperf_state=1
 for i in {1..240}; do
+  if [ "$(oc get benchmarks.ripsaw.cloudbulldozer.io -n my-ripsaw -o json | jq -r .items[].status.state)" == "Error" ]; then
```
chaitanyaenr, line 66: jq might or might not be installed, maybe leverage the jsonpath to capture the state: https://kubernetes.io/docs/reference/kubectl/jsonpath/. Thoughts?

mohit-sheth: yup, jsonpath++
updated

PR #38, opened 2020-08-06
have the gsheet key in a central location and rectified edge cases

mohit-sheth:
having the keys in central location helps when we are cloning the repo over the existing one ex: scale-ci-pipeline. Also we already have /root/.secrets dir for GCP install and it's service account key. so leveraging that and going forward this location would be good to store creds

rectified edge case
`touchstone_compare ... username:password@$_es:$_es_port     username:password@$_es:$_es_port ... `
instead of 
`touchstone_compare ... username:password@_es:_es_port         _es:_es_port   ... `
which resulted in auth error

### DIFF
--- a/workloads/network-perf/csv_gen.py
+++ b/workloads/network-perf/csv_gen.py
@@ -189,8 +189,8 @@ parser.add_argument(
 )
 params = parser.parse_args()
 generate_csv(params.yaml_files, f"{params.sheetname}.csv")
-if "EMAIL_ID_FOR_RESULTS_SHEET" in os.environ:
+if "EMAIL_ID_FOR_RESULTS_SHEET" and "GSHEET_KEY_LOCATION" in os.environ:
     push_to_gsheet(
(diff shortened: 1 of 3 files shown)


workloads/network-perf/csv_gen.py
```diff
@@ -191,6 +191,6 @@ def push_to_gsheet(csv_file_name, google_svc_acc_key, email_id):
 generate_csv(params.yaml_files, f"{params.sheetname}.csv")
 if "EMAIL_ID_FOR_RESULTS_SHEET" in os.environ:
     push_to_gsheet(
-        f"{params.sheetname}.csv", "gsheet_key.json", os.environ["EMAIL_ID_FOR_RESULTS_SHEET"],
+        f"{params.sheetname}.csv", "/root/.secrets/gsheet_key.json", os.environ["EMAIL_ID_FOR_RESULTS_SHEET"],
```
chaitanyaenr, line 194: Maybe change this to look for secrets under $HOME/.secrets instead of defaulting to root user, also being able to define the path to the secrets using a var/flag should make it more flexible for automation. Thoughts?

mohit-sheth: done

PR #41, opened 2020-09-03
add smoke test for uperf workload

mohit-sheth:
helps with testing the changes to the scale-ci-pipeline

### DIFF
--- a/workloads/network-perf/smoke_test.sh
+++ b/workloads/network-perf/smoke_test.sh
@@ -1,0 +1,118 @@
+#!/usr/bin/env bash
+set -x
+
+source ./common.sh
+
+if [[ ${COMPARE} == "true" ]]; then
+  _baseline_pod_1p_uuid=
+  _baseline_pod_2p_uuid=
+  _baseline_pod_4p_uuid=
+fi
+
+if [[ ${BASELINE_POD_1P_UUID} ]]; then
+  _baseline_pod_1p_uuid=${BASELINE_POD_1P_UUID}
+fi
+
+if [[ ${BASELINE_POD_2P_UUID} ]]; then
+  _baseline_pod_2p_uuid=${BASELINE_POD_2P_UUID}
+fi
+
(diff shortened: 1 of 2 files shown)


workloads/network-perf/smoke_test.sh
```diff
@@ -0,0 +1,118 @@
...
+      protos:
+        - tcp
+      test_types:
+        - stream
+      sizes:
+        - 1024
+      runtime: 30
+EOF
+
+sleep 30
```
chaitanyaenr, line 69: What's the need for this sleep?

mohit-sheth: it takes some time for the benchmark operator to come up

workloads/network-perf/smoke_test.sh
```diff
@@ -0,0 +1,118 @@
...
+fi
+
+if [[ ${BASELINE_POD_2P_UUID} ]]; then
+  _baseline_pod_2p_uuid=${BASELINE_POD_2P_UUID}
+fi
+
+if [[ ${BASELINE_POD_4P_UUID} ]]; then
+  _baseline_pod_4p_uuid=${BASELINE_POD_4P_UUID}
+fi
+
```
jtaleric, line 23: Seems like all the above could be removed? Since we won't be comparing in the smoke test, right?

mohit-sheth: so the smoke test's primary reason is to test the PRs from scale-ci-pipeline. 
Pipeline basically git clones and runs either hn2hn, pod or service which will take 1 hour each at best. The parameters inside like runtime, message sizes, protos are not configurable from the pipeline perspective.
Hence this smoke test will be a quick way to do e2e tests for PR

jtaleric: ack

jtaleric (approved):
LGTM

chaitanyaenr (approved):
LGTM

PR #44, opened 2020-09-23
fix path for yaml files; add ns arg for query

mohit-sheth:
compare.yaml files will be created in /network-perf instead of /network-perf/touchstone as we are installing touchstone with pip

### DIFF
--- a/workloads/network-perf/run_network_compare.sh
+++ b/workloads/network-perf/run_network_compare.sh
@@ -39,7 +39,7 @@ if [[ $? -ne 0 ]] ; then
   exit 1
 fi
 set -x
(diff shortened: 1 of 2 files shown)


jtaleric (approved):
LGTM

rsevilla87:
Hey @mohit-sheth , this line also needs the -n flag
https://github.com/cloud-bulldozer/e2e-benchmarking/blob/fabd5cbbc24e5217978262e20718bf13ed1a9455/workloads/router-perf/run_router_test.sh#L45

rsevilla87 (changes requested):
Added some comments

rsevilla87 (approved):
LGTM!

PR #46, opened 2020-09-28
Add kube-burner e2e test

rsevilla87:
Added a mechanism to label nodes and calculate number of pods to deploy for kubelet density tests.


workloads/kube-burner/README.md
```diff
@@ -0,0 +1,35 @@
...
+| **ES_INDEX**         | Elastic search index            | ripsaw-kube-burner|
+| **PROM_URL**         | Elastic search endpoint         | https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091|
+| **JOB_TIMEOUT**      | kube-burner's job timeout       | 7200|
+| **WORKLOAD_NODE**    | Workload node name              | "" (don't pin kube-burner to any server)|
+| **TAINT_NODE**       | Taint the workload node         | 0 (don't taint)|
+| **CERBERUS_URL**     | URL to check the health of the cluster using [Cerberus](https://github.com/openshift-scale/cerberus) | "" (don't check)|
+
+### cluster-density variables
+
+The `cluster-density` workload supports the environment variable **JOB_ITERATIONS**. This variable configures the number of cluster-density jobs iterations to perform (1 namespace per iteration). By default 1000
```
chaitanyaenr, line 28: We might want to mention the number of objects which get created in each namespace to be able to estimate the load cluster will be able to handle and set the number of namespaces accordingly. Thoughts?

workloads/kube-burner/common.sh
```diff
@@ -0,0 +1,117 @@
...
+}
+
+deploy_workload() {
+  log "Deploying benchmark"
+  envsubst < kube-burner-crd.yaml | oc apply -f -
+}
+
+wait_for_benchmark() {
+  rc=0
+  log "Waiting for benchmark kube-burner-${1}-${UUID} to complete"
```
chaitanyaenr, line 68: We might want to stream the logs of the kube-burner pods to stdout. It will be very useful in case of jenkins to see the progress. Thoughts?

rsevilla87: nice, I like it

jtaleric (commented):
Can we include a env.sh file that allows us to run this outside of a CI system ?

rsevilla87:
> Can we include a env.sh file that allows us to run this outside of a CI system ?

Done

jtaleric (commented):
What if the ES server needs auth?

workloads/kube-burner/common.sh
```diff
@@ -0,0 +1,124 @@
+
+# Check cluster's health
+if [[ ${CERBERUS_URL} ]]; then
```
chaitanyaenr, line 3: We might want to add a Cerberus check during/at the end of the run as well to make sure the upgraded cluster is healthy. We can query ripsaw as well, for example: https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/network-perf/run_pod_network_test_fromgit.sh#L77. Think the benchmark will fail if Cerberus reports issues right? In that case we don't need an explicit check. Thoughts?

jtaleric: Why do we need to check this, if the Operator does? seems unnecessary.

rsevilla87: @chaitanyaenr instead trusting in the curl,  I added the parameter cerberus_url to the CRD so now ripsaw will take care of checking cerberus status.

workloads/kube-burner/kube-burner-crd.yaml
```diff
@@ -0,0 +1,71 @@
...
+      # Use a custom kube-burner image
+      image: quay.io/cloud-bulldozer/kube-burner:latest
+      # Queries per second
+      qps: ${QPS}
+      # Max number of burst queries to perform
+      burst: ${BURST}
+      # Log level. Allowed, info and debug
+      log_level: info
+      # Delete old namespaces for the selected workload before starting benchmark
+      cleanup: false
```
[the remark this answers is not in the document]
rsevilla87: :+1:

workloads/kube-burner/common.sh
```diff
@@ -0,0 +1,133 @@
...
+  log "Removing kubelet-density=enabled label from worker nodes"
+  for n in ${nodes}; do
+    oc label ${n} kubelet-density-
+  done
+}
+
+check_running_benchmarks() {
+  benchmarks=$(oc get benchmark -n my-ripsaw --ignore-not-found | wc -l)
+  if [[ ${benchmarks} -gt 0 ]]; then
+    echo ${benchmarks} | grep kube-burner | grep -qE "Failed|Completed" || log "Another kube-burner benchmark is running at the moment" && exit 1
```
chaitanyaenr, line 131: Think we need to grep for benchmarks in "Running" state?

rsevilla87: There's a period of time a benchmark doesn't have state. Grep-ing for Failed and Complete is the best way to be sure there's no benchmark running at the moment

workloads/kube-burner/README.md
```diff
@@ -0,0 +1,59 @@
...
+- kubelet-density-heavy
+
+## Environment variables
+
+All scripts can be tweaked with the following environment variables:
+
+| Variable         | Description                         | Default |
+|------------------|-------------------------------------|---------|
+| **QPS**              | Queries/sec                     | 10      |
+| **Burst**            | Burst queries                   | 10      |
```
chaitanyaenr, line 16: s/Burst/BURST

rsevilla87: I think not. Official docs say Burst

rsevilla87: good catch!, fixed

chaitanyaenr (approved):
LGTM

jtaleric (approved):
LGTM!

PR #51, opened 2020-10-02
switch order of uuid

mohit-sheth:
A bug that surfaced when the baseline elasticsearch instance and current run's elasticsearch instance are different


rsevilla87 (approved):
LGTM

PR #52, opened 2020-10-05
Avoid deleting the benchmark to track the status in CI

chaitanyaenr:
This commit modifies the e2e scripts to avoid deleting the benchmark
at the end of the run to be able to capture the status of the run in
CI. The benchmark is cleaned up before kicking off a new run instead.

### DIFF
--- a/workloads/network-perf/run_multus_network_tests_fromgit.sh
+++ b/workloads/network-perf/run_multus_network_tests_fromgit.sh
@@ -18,8 +18,7 @@ if [[ ${PAIR} ]]; then
 fi
 
 if ${MULTUS} ; then
-oc -n my-ripsaw delete benchmark/uperf-benchmark
-
+oc -n my-ripsaw delete benchmark/uperf-benchmark-multus-network --wait
 if [[ ${MULTUS_SERVER_NAD} ]]; then
   MULTUS_SERVER="server: ${MULTUS_SERVER_NAD}"
 fi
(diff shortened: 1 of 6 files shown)


workloads/network-perf/run_multus_network_tests_fromgit.sh
```diff
@@ -26,7 +26,8 @@ if [[ ${PAIR} ]]; then
 fi
 
 if ${MULTUS} ; then
-oc -n my-ripsaw delete benchmark/uperf-benchmark
+oc -n my-ripsaw delete benchmark/uperf-benchmark-multus-network --wait
+fi
```
dry923, line 30: This fi shouldn't be here. the if matches to the fi from line 123

dry923: It also begs the question as to why we have the if there to being w/ given this is the multus test. You wouldn't be running it if you didn't want it. Thoughts on getting rid of the surrounding if?

chaitanyaenr: Makes sense @dry923, removed it.

mohit-sheth:
we might have to query the benchmark with it's name here https://github.com/cloud-bulldozer/e2e-benchmarking/blob/073af74911ff03bd3150d1597ba15497daa89d6e/workloads/network-perf/run_pod_network_test_fromgit.sh#L77

whitleykeith (approved):

chaitanyaenr:
> we might have to query the benchmark with it's name here
> 
> https://github.com/cloud-bulldozer/e2e-benchmarking/blob/073af74911ff03bd3150d1597ba15497daa89d6e/workloads/network-perf/run_pod_network_test_fromgit.sh#L77
Fixed it, thanks.

mohit-sheth (approved):
LGTM

PR #53, opened 2020-10-06
Add upgrade test to e2e

dry923:
Add an upgrade test to the e2e scripts. 

Depends-on: https://github.com/cloud-bulldozer/benchmark-wrapper/pull/216
Until ^ is merged it is targeting my repo for testing.

### DIFF
--- a/workloads/upgrade-perf/README.md
+++ b/workloads/upgrade-perf/README.md
@@ -1,8 +1,7 @@
-# Scale Scripts
+# Upgrade Scripts
 
-The purpose of the scale scripts is scale Openshift Cluster to a given size.
-It will scale up and down X number of times and report the timings to 
-Elasticsearch
+The purpose of the upgrade scripts is to upgrade an Openshift Cluster
+to a given version. It will report the timings to Elasticsearch.
 
 Running from CLI:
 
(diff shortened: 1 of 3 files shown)


workloads/upgrade-perf/common.sh
```diff
@@ -0,0 +1,93 @@
...
+else
+  echo $cloud_name > uuid.txt
+fi
+
+echo "Installing snafu in python virtual environment"
+
+rm -rf /tmp/snafu upgrade
+
+# REPLACE ONCE WRAPPER IS MERGED
+git clone https://github.com/dry923/snafu.git --branch upgrade /tmp/snafu
```
mohit-sheth, line 74: upstream:master before merge

workloads/upgrade-perf/common.sh
```diff
@@ -0,0 +1,93 @@
...
+
+python3 -m venv upgrade
+source upgrade/bin/activate
+pip3 install -e /tmp/snafu
+
+es_index="openshift-upgrade-timings"
+es=${_es}
+es_port=${_es_port}
```
mohit-sheth, line 82: where are these variables used?

workloads/upgrade-perf/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+### ES_SERVER
+Default: `milton.aws.com`  
+Elasticsearch server to index the results of the current run
+
+### ES_PORT
+Default: ``  
+Port number for elasticsearch server
+
+### COMPARE (not implemented yet)
```
chaitanyaenr, line 47: We might want to remove the uperf reference.

workloads/upgrade-perf/common.sh
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
+set -x
+
+# Check cluster's health
+if [[ ${CERBERUS_URL} ]]; then
```
chaitanyaenr, line 5: We might want to add a Cerberus check at the end of the run as well to make sure the upgraded cluster is healthy. We can query ripsaw as well, for example: https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/network-perf/run_pod_network_test_fromgit.sh#L77. Thoughts?

chaitanyaenr: Think the benchmark will fail if Cerberus reports issues right? In that case we don't need an explicit check.

workloads/upgrade-perf/common.sh
dry923: yup. Still waiting on the wrapper merge then i'll change it.

workloads/upgrade-perf/common.sh
dry923: snafu expects them to be defined in the environment. Not sure why it cant just accept cmnd line args but thats a different discussion really

workloads/upgrade-perf/common.sh
mohit-sheth: ah, missed the description
my bad

workloads/upgrade-perf/README.md
dry923: HA! Good catch. Also fixed in the in scale doc since its a copy of a copy.

mohit-sheth:
LGTM

mohit-sheth (approved):

chaitanyaenr (approved):
LGTM

PR #55, opened 2020-10-09
Setting metadata collection to be done via daemonset for serviceip

dry923:
The ServiceIP test was having issues due to the restricted port range. When metadata collection was done the Elasticsearch ports were not being released so port 20000 was frequently in use and would cause uperf to fail.


mohit-sheth:
LGTM

mohit-sheth (approved):

jtaleric (approved):
LGTM!

PR #58, opened 2020-10-21
Add a current/desired worker count status message

dry923:
Adding a status message so we know what the scale is currently at and what is desired.

example output:
```
...
+ echo 'Current worker count: 3'
Current worker count: 3
+ echo 'Desired worker count: 4'
Desired worker count: 4
...
```


mohit-sheth (approved):
LGTM

PR #63, opened 2020-10-27
consume gold-index for comparison

mohit-sheth:
With this commit we can use gold-index for comparison. We can also manually add the uuids which will override the gold-index uuids. 
If we want to compare against gold, we need to specify the ocp version to compare against and the sdn type to compare against (defaults to cluster sdn)
Also, we need to specify the elasticsearch server where the gold-index is housed

### DIFF
--- a/workloads/network-perf/README.md
+++ b/workloads/network-perf/README.md
@@ -17,19 +17,19 @@ $ ./run_<test-name>_network_test_fromgit.sh
 ## Environment variables
 
 ### ES_USER
-Default: `` 
+Default: ``    
 Username for elasticsearch instance
 
 ### ES_PASSWORD
-Default: `` 
+Default: ``     
 Password for elasticsearch instance
 
 ### ES_SERVER
-Default: `milton.aws.com`  
+Default: `milton.aws.com`    
 Elasticsearch server to index the results of the current run
 
 ### ES_PORT
-Default: ``  
+Default: ``    
 Port number for elasticsearch server
 
 ### METADATA_COLLECTION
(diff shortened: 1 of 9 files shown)


workloads/network-perf/README.md
```diff
@@ -17,35 +17,52 @@ $ ./run_<test-name>_network_test_fromgit.sh
...
+Default: ``    
 Port number for elasticsearch server
 
 ### METADATA_COLLECTION
 Default: `true`   
 Enable/Disable collection of metadata
 
 ### COMPARE
-Default: `false`   
+Default: `false`
```
jtaleric, line 40: Something happened here... you added random spaces?

workloads/network-perf/README.md
```diff
@@ -17,35 +17,52 @@ $ ./run_<test-name>_network_test_fromgit.sh
...
+Default: `false`        
 Enable/Disable the ability to compare two uperf runs. If set to `true`, the next set of environment variables pertaining to the type of test are required
 
+### COMPARE_WITH_GOLD
+Default: ``     
+If COMPARE is set to true and COMPARE_WITH_GOLD is set to true then the current run will be compared against our gold-index
+Note: Make sure that elasticsearch baseline uuid (example: BASELINE_POD_1P_UUID, BASELINE_POD_2P_UUID ...) vars are not set or else it will override the uuids
+
+### GOLD_SDN
+Default: `you current cluster's sdn`
```
jtaleric, line 49: Lets just default this to openshiftsdn for now

workloads/network-perf/README.md
mohit-sheth: The next line continues after `false` if we don't add enough spaces

workloads/network-perf/README.md
mohit-sheth: done

jtaleric (approved):
LGTM - @mohit-sheth assuming you tested this locally?

PR #65, opened 2020-10-29
kube-burner log streaming fix

rsevilla87:
When job finishes, the log_streaming loop iterates several times (and therefore printing duplicated logs) before exiting because the benchmark hasn't been marked as Completed|Failed yet. This adds an extra time period to let benchmark-operator update it on time.


chaitanyaenr (approved):
LGTM

PR #70, opened 2020-11-17
Ignore all log streming errors and modify default metrics profiles

rsevilla87:
Fatal errors are not ignored by the  --ignore-errors flag, adding `|| true` will ignore all errors now.


jtaleric (approved):
lgtm

jtaleric (approved):
LGTM! Are we seeing the jobs take longer, hence the time change?

rsevilla87:
> bs take longer, hence the time change

Yes, I've seen cluster-density 4K taking more that 4 hours.

PR #75, opened 2020-11-24
Fix running kube-burner benchmark

rsevilla87:
The current implementation doesn't verify if a running benchmark is of the kind kube-burner.


chaitanyaenr (approved):
LGTM

PR #78, opened 2020-11-30
Add export to fix indexing issue

dry923:
The upgrade was not properly passing the ES index/port/server information to snafu. Adding the export in front of the variable definition resolves the issue.


chaitanyaenr (approved):
LGTM

PR #82, opened 2020-12-03
Apply the backpack_role.yaml

dry923:
The backpack_role.yaml was no longer being created after a recent refactoring and would cause a test to fail because it cannot find the backpack-view service account. This adds the apply back in.


rsevilla87 (approved):
LGTM!

mohit-sheth (approved):
LGTM

PR #92, opened 2021-01-07
Compare and gen CSV only when COMPARE is true

rsevilla87:
The current implementation doesn't allow to run uperf w/o comparing or generating CSV, we should to be able to generate a CSV w/o needing a baseline UUID in the future. That should  be addressed in a different patch.

### DIFF
--- a/workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
+++ b/workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
@@ -9,7 +9,7 @@ wait_for_benchmark
 assign_uuid
 delete_benchmark
 if [[ ${COMPARE} == "true" ]]; then
-  run_benchmark_comparison
-  generate_csv
+  run_benchmark_comparison compare_output_${pairs}.yaml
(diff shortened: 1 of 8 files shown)


workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
```diff
@@ -7,7 +7,9 @@ export pairs=1
 deploy_workload
 wait_for_benchmark
 assign_uuid
-run_benchmark_comparison
 delete_benchmark
+if [[ ${COMPARE} == "true" ]]; then
```
mohit-sheth, line 11: we need to generate csv even when COMPARE=false
it will just create current run's results csv

rsevilla87: Makes sense, will implement the change.

mohit-sheth:
we can remove smoke-test.sh completely
instead we can create a new smoke-test specific crd and apply it when SMOKE_TEST=true (MOAR VARSS). Thoughts @rsevilla87?
might also need a small change in readme

PR #93, opened 2021-01-07
have a default run_id set to NA for non pipeline runs

mohit-sheth:
avoids 
```
usage: run_snafu [-v] -t TOOL [--run-id RUN_ID]
run_snafu: error: argument --run-id: expected one argument
OK
```
when you run it directly from e2e and not from pipeline


jtaleric (changes requested):
One sec... I don't think this is the right change.

jtaleric:
https://github.com/cloud-bulldozer/benchmark-wrapper/pull/241

mohit-sheth:
closing in favour of ^^

PR #102, opened 2021-01-20
add metadata.targeted; set it false for serviceip

mohit-sheth:
there have been clashes for the limited open ports (20000-20011)
this commit enables backpack daemonset mode for serviceip so that we don't see any clashes


jtaleric (approved):
LGTM!

PR #104, opened 2021-01-27
Pod density and workload summary

rsevilla87:
Executing a pod density workload using node-density could results tricky. Having an specific script for this workload will make it simpler.

```bash
$ export PODS=1000
$ ./run_poddensity_test_fromgit.sh
```
Also adding some workload information to the log, that can be useful to verify variables values before triggering the actual workload.


smalleni:
@rsevilla87 My only concern with this is that the way it is right now, the namespace will still be name node-density-{{uuid}} right? Just a nit...but still

rsevilla87:
> @rsevilla87 My only concern with this is that the way it is right now, the namespace will still be name node-density-{{uuid}} right? Just a nit...but still

Yeah, you're right. We can merge you implementation instead.

smalleni:
> > @rsevilla87 My only concern with this is that the way it is right now, the namespace will still be name node-density-{{uuid}} right? Just a nit...but still
> 
> Yeah, you're right. We can merge you implementation instead.

Cool, review on https://github.com/cloud-bulldozer/benchmark-operator/pull/528 then please and then https://github.com/cloud-bulldozer/e2e-benchmarking/pull/123

PR #111, opened 2021-02-04
Substract kube-burner pod from the node-density pod count

rsevilla87:
We're not counting the kube-burner pod itself when calculating the number of pods to deploy. This is generating issues since this pod can't be scheduled, thus kube-burner times out waiting for it to be ready.


rsevilla87:
@chaitanyaenr as we talked internally, I've updated the WORKLOAD_NODE variable to use noseSelector rather than node hostname

chaitanyaenr (approved):
LGTM

Issue #60, opened 2020-10-22
CI pipeline


rsevilla87:
We need a CI since these scripts are becoming more critical.

whitleykeith:
Jsst a few questions

How do we consume these scripts? Do teams cherry pick these when they want to run them in their own CI, are they part of the larger builds etc.

If we're targeting these for consumption, we should think through how we would want to surface the core logic. What do we want the artifact to look like? How do we handle the large amount of parameters that need to be set currently? Do we care about surfacing these up to a more general audience, say general Kubernetes clusters?

jtaleric:
> Just a few questions
> 
> How do we consume these scripts? Do teams cherry pick these when they want to run them in their own CI, are they part of the larger builds etc.
> 
> If we're targeting these for consumption, we should think through how we would want to surface the core logic. What do we want the artifact to look like? How do we handle the large amount of parameters that need to be set currently? Do we care about surfacing these up to a more general audience, say general Kubernetes clusters?

Ah, just saw this, so let me address this question... 

> How do we consume these scripts?

Today, we consume these scripts in our Performance CI - we define the environmental vars and call the specific e2e-benchmarking workload. 

We would like to eventually have other teams use/contribute....

rsevilla87:
Fixed by #114

PR #114, opened 2021-02-08
Created CI pipeline for e2e-benchmarking

Harshith-umesh:


README.md
```diff
@@ -1,3 +1,7 @@
 # End to End Benchmarking Scripts
 This repo contains performance test scripts which install, configure, execute and evaluate workloads.
 
+CI command to run all tests : ./ci/run_ci.sh
```
jtaleric, line 4: I would move this to a `README` in `/ci`.

ci/env.sh
```diff
@@ -0,0 +1,21 @@
+#smoke test environment specifications:
+
+#for scale perf
+export SCALE=4
+export RUNS=1
+
+#for kube burner 
+export JOB_ITERATIONS=3
```
jtaleric, line 8: Do we need 3 iterations for the test?

chaitanyaenr: We define the number of projects to create using the JOB_ITERATIONS var.

ci/env.sh
```diff
@@ -0,0 +1,21 @@
...
+export CLEANUP=true
+export CLEANUP_WHEN_FINISH=true
+export NODE_COUNT=4
```
chaitanyaenr, line 11: This depends on the number of the nodes running in the cluster on which the test is being run, maybe set this is to the node count present in the cluster being used by benchmark-wrapper? Thoughts?

ci/run_ci.sh
```diff
@@ -0,0 +1,69 @@
...
+  directory=${test%:*}                           #to extract the workload directory name 
+  echo $directory  $command
+
+  figlet "CI test for ${test}"                   # figlet gives a heading for each test
+
+  cd workloads/
+  cd $directory
+  echo $PWD
+
+  (sleep 900s; sudo pkill -f $command) &      #to kill a workload script if it doesn't execute within 500s ; runs in the background
```
chaitanyaenr, line 46: We might want to make this configurable with a default to make sure we cover the test cases which might take longer than the specified duration in the future? Thoughts?

ci/CI_test_list.yml
```diff
@@ -0,0 +1,12 @@
...
+kube-burner:run_clusterdensity_test_fromgit.sh
+kube-burner:run_maxnamespaces_test_fromgit.sh 
+kube-burner:run_maxservices_test_fromgit.sh 
+kube-burner:run_nodedensity_test_fromgit.sh 
+kube-burner:run_nodedensity-heavy_test_fromgit.sh 
+network-perf:smoke_test.sh 
+storage-perf:run_storage_tests_fromgit.sh 
+upgrade-perf:run_upgrade_fromgit.sh 
+router-perf:run_router_test.sh 
+
```
chaitanyaenr, line 12: Let's run the scale workload again in the end to get the number of nodes back to the previous count before the scale-perf workload ran.

chaitanyaenr (commented):
Minor nits

README.md
Harshith-umesh: yes makes sense , will do the change

ci/run_ci.sh
Harshith-umesh: yes makes sense , will add an env var named TEST_TIMEOUT in env.sh which can be initialized with a default value of 10 or 15 mins .

ci/CI_test_list.yml
Harshith-umesh: yes we should do that , i'll make the changes

ci/env.sh
Harshith-umesh: yes we could do that

ci/run_ci.sh
```diff
@@ -0,0 +1,79 @@
...
+
+for test in ${test_list}; do            
+
+  start_time=`date`
+
+  command=${test##*:}                                     #to extract the shell script name to run
+  directory=${test%:*}                                    #to extract the workload directory name 
+  echo $directory  $command
+
+  figlet "CI test for ${test}"                            # figlet gives a heading for each test
```
chaitanyaenr, line 41: Maybe use echo instead of figlet to avoid dependency on the package? Thoughts?

chaitanyaenr (approved):
LGTM, nice work @Harshith-umesh.

ci/run_ci.sh
```diff
@@ -0,0 +1,80 @@
...
+  start_time=`date`
+
+  command=${test##*:}                                     #to extract the shell script name to run
+  directory=${test%:*}                                    #to extract the workload directory name 
+  
+  echo -e "\n======================================================================"
+  echo -e "     CI test for ${test}                    "
+  echo -e "======================================================================\n"
+
+  if [[ ( $directory == scale-perf ) && ( $bit == 0 ) ]]; #check to increment number of worker nodes by 1 and to bring it
```
amitsagtani97, line 44: when user will try to run only scale-test using "/rerun scale-perf/........" the scale-down will not be triggered.

Instead, check the number of worker node at the end of all / single test and if it is not equal to $ORIGINAL_WORKER_COUNT then run the scale job with SCALE=$ORIGINAL_WORKER_COUNT

amitsagtani97 (changes requested):

amitsagtani97 (approved):
nice work @Harshith-umesh 
LGTM!

ci/env.sh
```diff
@@ -0,0 +1,26 @@
+#smoke test environment specifications:
```
chaitanyaenr, line 1: We might want to add a way to specify kubeconfig path, it's useful when the kubeconfig is not under the default location - $HOME/.kube/config. Something like KUBECONFIG_PATH, this way we can run export KUBECONFIG=$KUBECONFIG_PATH command in the script for the oc client to pick it up and talk to the cluster. Thoughts?

The workloads need to be modified to support it as well but it might be out of the scope for this PR. See https://github.com/cloud-bulldozer/e2e-benchmarking/issues/118.

mohit-sheth (approved):
LGTM

ci/env.sh
```diff
@@ -0,0 +1,31 @@
...
+export SCALE=$NEW_WORKER_COUNT
+export RUNS=1
+
+#for kube burner 
+export JOB_ITERATIONS=3
+export CLEANUP=true
+export CLEANUP_WHEN_FINISH=true
+export NODE_COUNT=`oc get nodes -l node-role.kubernetes.io/worker | grep -v NAME | wc -l`
+export PODS_PER_NODE=40
+if [[ -n $KUBECONFIG_PATH ]]; then
```
chaitanyaenr, line 16: We might want to document this variable in the readme for the users to leverage it when using a kubeconfig which is not in the default location.

Harshith-umesh: yes makes sense , will add it right away

amitsagtani97:
Fixes #60

PR #115, opened 2021-02-08
Fix disable ES indexing

rsevilla87:
The current implementation does not allow to disable ES indexing. Fixing it.


chaitanyaenr (approved):
LGTM

PR #117, opened 2021-02-12
Fix conditional

rsevilla87:
The conditional was not working well, updating it.


dry923 (approved):
LGTM!

PR #135, opened 2021-03-05
added a node to scale option , default worker

Harshith-umesh:


workloads/scale-perf/run_scale_fromgit.sh
```diff
@@ -44,7 +44,7 @@ spec:
     name: scale_openshift
     args:
       label:
-        key: node-role.kubernetes.io/workload
+        key: node-role.kubernetes.io/${NODE_TO_SCALE:=worker}
```
rsevilla87, line 47: We usually set these variables in common.sh. Also I think is better to use WORKLOAD_NODE_ROLE rather than NODE_TO_SCALE

Harshith-umesh: yes makes sense , will do the needful

rsevilla87 (approved):
LGTM!

PR #141, opened 2021-03-17
Provide prereqs for Python code for router-perf-v2

### LINKED ISSUE #140: Provide a requirements.txt for workload/router-perf-v2
workload and csv_gen use some uncommon pre-reqs.  Provide and document an easy way to install them to avoid a nasty surprise with csv_gen failing after a long run.

mffiedler:
Provide an easy way to install pre-reqs for workload.py and csv_gen.py.

Additionally, update csv_gen.py to not rely on system python3 in case user is running in a virtualenv

Fixes #140 

@rsevilla87 @mohit-sheth PTAL

### DIFF
--- a/workloads/router-perf-v2/README.md
+++ b/workloads/router-perf-v2/README.md
@@ -27,6 +27,7 @@ Apart from the k8s/oc clients, running this script has several requirements:
 
 - jq
 - podman
+- pip install -r requirements.txt
 
 ## Configuration
 It's possible to tune the default configuration through environment variables. They are described in the table below:
--- a/workloads/router-perf-v2/csv_gen.py
+++ b/workloads/router-perf-v2/csv_gen.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 import yaml
 from collections import defaultdict
 import csv
--- a/workloads/router-perf-v2/requirements.txt
+++ b/workloads/router-perf-v2/requirements.txt
@@ -1,0 +1,6 @@
+PyYAML
+oauth2client>=4.1.3
+gspread
+gspread-formatting
+numpy
+elasticsearch


rsevilla87:
Maybe we can execute `pip install -r requirements.txt` from the code?

mffiedler:
> Maybe we can execute `pip install -r requirements.txt` from the code?

Good idea,  I am thinking before deploy_infra.  Let me know if there is a better spot.  Will test that on a clean virtualenv and add a new commit.

rsevilla87:
> > Maybe we can execute `pip install -r requirements.txt` from the code?
> 
> Good idea, I am thinking before deploy_infra. Let me know if there is a better spot. Will test that on a clean virtualenv and add a new commit.

Yeah, that would be a perfect spot.

kedark3:
> > Maybe we can execute `pip install -r requirements.txt` from the code?
> 
> Good idea, I am thinking before deploy_infra. Let me know if there is a better spot. Will test that on a clean virtualenv and add a new commit.

if we put that in the code, wouldn't it also make sense to create an ephemeral virtualenv? We wouldn't want to try to install some pip packages on user's system, and likely to fail.

mffiedler:
I think @kedark3 is correct.   Least surprise would be letting the end-user fulfill the pre-reqs themselves unless we want to get into the business of managing virtualenvs which we do not do for our other tools.

rsevilla87:
We're already using venvs in other tools:

https://github.com/cloud-bulldozer/e2e-benchmarking/blob/bb226b0acbc2483e053528a14678d6a04ee0db3c/workloads/upgrade-perf/common.sh#L58-L60
https://github.com/cloud-bulldozer/e2e-benchmarking/blob/fe879cac4821ea569be8ee2cd7a70ccbfe428adb/utils/touchstone-compare/run_compare.sh#L7-L9

mohit-sheth (approved):
LGTM

rsevilla87 (approved):
LGTM

PR #142, opened 2021-03-17
Generate large-scale and small-scale scenarios

rsevilla87:
The current CI implementation does not allow to execute the same job with different variables w/o tricks. We need to cover small and large scale scenarios. This PR splits the current implementation those two, hence, we'll need to perform some changes in the scale-ci-pipeline code.

Apart from this change. There were several mix termination scenarios we were not covering.
- In small scale mix termination is tested with these numbers
40 routes with 1, 40, 200 threads
400 routes with 1, 20, 80 threads

- In large scale mix termination is tested with these numbers:
200 routes with 1, 20, 80 threads
2000 routes with 1, 10, 20 threads

In addition to the above:
- Remove the RAMP_UP variable since It does not work well, mb generates logs even during this ramp-up period, that lead to get wrong results.
- Incrementing default QUIET_PERIOD to 60s (timeout of TIME_WAIT sockets)
- Decreasing default number of SAMPLES to 1


workloads/router-perf-v2/README.md
```diff
@@ -19,13 +19,26 @@ for clients in CLIENTS; do
 done
 ```
 
+This folter contains two different variants:
```
mffiedler, line 22: Typo:  folder

kedark3, line 22: `s/folter/folder/` ?

workloads/router-perf-v2/ingress-performance-large.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/bash -e
+set -e
+
+. common.sh
+export NUMBER_OF_ROUTES=${NUMBER_OF_ROUTES:-500}
+CLIENTS="1 20 80"
```
mffiedler, line 6: Why override clients?   Can we default it to this value or accept a value set by the user?  I can see where being able to configure it would be useful - e.g. to shorten a run to validate the test is working well.

workloads/router-perf-v2/README.md
```diff
@@ -39,16 +52,13 @@ It's possible to tune the default configuration through environment variables. T
 | RUNTIME				| Workload duration in seconds | `120` |
 | TERMINATIONS  		| List of HTTP terminations to test | `http edge passthrough reencrypt mix` |
 | URL_PATH              | URL path to use in the benchmark | `/1024.html` |
-| NUMBER_OF_ROUTES		| Number of routes of each termination to test. (the mix scenario test all of them together) | `100` |
```
kedark3, line 42: out of curiosity, why are these parameters being removed? Do we no longer expect user to be able to configure these items? I feel like leaving it configurable maybe still valuable for us? Thoughts @mffiedler ?

mffiedler (changes requested):

workloads/router-perf-v2/README.md
```diff
@@ -39,16 +52,13 @@ It's possible to tune the default configuration through environment variables. T
...
-| CLIENTS				| List with the number of clients (threads per target route) | `1 10 100`|
 | TLS_REUSE				| Reuse TLS session | `true` |
-| RAMP_UP				| Ramp-up period | `0 ` |
-| SAMPLES				| Number of samples to perform of each test | `3` |
+| SAMPLES				| Number of samples to perform of each test | `1` |
 | HOST_NETWORK			| Enable hostNetwork in the mb client | `true` |
 | NUMBER_OF_ROUTERS		| Number of routers to test | `2` |
 | NODE_SELECTOR			| Node selector of the mb client | `{node-role.kubernetes.io/workload: }` |
-| QUIET_PERIOD			| Quiet period after each test iteration | `10s` |
+| QUIET_PERIOD			| Quiet period after each test iteration | `60s` |
```
kedark3, line 61: :+1:

mffiedler:
General comment:  Would it be cleaner to have an env_large.sh and env_small.sh with parms customized for the type of run?   That way the end user can have env_mine.sh and still run the same base script.

rsevilla87:
Closing this PR in favor of https://github.com/cloud-bulldozer/e2e-benchmarking/pull/144

PR #144, opened 2021-03-18
Large-scale and small-scale scenarios for router-perf

rsevilla87:
After an internal conversation with @mohit-sheth, we've decided to make some changes in the workload. So the results we get can represent better a real case scenario:

Important changes:

- The script is now able to detect large scale and small scale scenarios. The number of clients and routes used by these scenarios are defined in different variables. This will allow us to have a single entrypoint (this script) to execute small-scale or large-scale scenarios w/o needing to have 2 different scripts. This is required because our current CI pipeline implementation does not allow to execute the same job multiple times with different variables each.
- Default RUNTIME has been dropped to 60 seconds
- Default SAMPLES has been increased to 2
- Default clients for the small scale scenario (100 routes) are now: 1 40 200 and 1 20 80 for the mix termination
- Default clients for the large scale scenario (500 routes) are now:  1 20 80 and 1 10 20 for the mix termination
- Keepalive requests has been updated to: 0 1 50. (where 0 means to keepalive each connection during the test duration) 
, 1 is representative for scenarios representing ephemeral connections (i.e IoT devices sending updates). And 50 is representative for a normal user web browsing.

All these parameters can be tweaked, however our plan is to use these defaults.

cc: @chaitanyaenr @rook @mffiedler @dry923 @kedark3


kedark3 (approved):
PR looks good to me.

kedark3:
> Pending: This implementation only allows comparing an UUID against BASELINE_UUID. I thinking of splitting this variable in two: LARGE_SCALE_BASELINE_UUID and SMALL_SCALE_BASELINE_UUID. Thoughts?

I think this might be a good idea. It will help us see how our test compares to different baselines and make better analysis/decisions.

mohit-sheth:
>Default RUNTIME has been dropped to 60 seconds
>Default SAMPLES has been increased to 2

nice

>Default clients for the small scale scenario (100 routes) are now: 1 40 200 and 1 20 80 for the mix termination
>Default clients for the large scale scenario (500 routes) are now: 1 20 80 and 1 10 20 for the mix termination

we can get new gold uuids for 4.7 run with ^ scenario

mohit-sheth (approved):
LGTM

PR #145, opened 2021-03-22
Add more checks in the workloads scripts

amitsagtani97:
# Description
Adding more checks for confirmation of pod creation and running in workload scripts for better error debugging.

### DIFF
--- a/utils/common.sh
+++ b/utils/common.sh
@@ -8,7 +8,7 @@ function get_pod () {
   pod_name="False"
   until [ $pod_name != "False" ] ; do
     sleep $sleep_time
-    pod_name=$(kubectl get pods -l $1 --namespace ${3:-my-ripsaw} -o name | cut -d/ -f2)
+    pod_name=$(oc get pods -l $1 --namespace ${3:-my-ripsaw} -o name | cut -d/ -f2)
     if [ -z $pod_name ]; then
       pod_name="False"
     fi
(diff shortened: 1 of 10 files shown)


mohit-sheth (approved):
LGTM

workloads/scale-perf/run_scale_fromgit.sh
```diff
@@ -55,19 +56,38 @@ spec:
...
+      if [ $? -ne 0 ]; 
+      then 
+        exit 1
+      fi
+
+      uuid=${long_uuid:0:8}
 
-      sleep 30
+      # Checks the presence of scale pod. Should exit if pod is not available.
+      scale_pod=$(get_pod "app=scale-$uuid" 300)
```
chaitanyaenr, line 69: Maybe also check if the pod is in running state?

utils/common.sh
```diff
@@ -0,0 +1,54 @@
...
+  done
+  echo $uuid
+  return 0
+}
+
+# The argument is 'timeout and 'pod name' in seconds'
+function check_pod_ready_state () {
+  pod_name=$1
+  timeout=$2
+  kubectl wait --for=condition=ready pods --namespace ${3:-my-ripsaw} $pod_name --timeout=$timeout
```
chaitanyaenr, line 52: Maybe use oc instead of kubectl to be consistent? Thoughts?

utils/common.sh
```diff
@@ -0,0 +1,54 @@
...
+      echo "Unable to fetch the benchmark uuid!"
+      return 1
+    fi
+  done
+  echo $uuid
+  return 0
+}
+
+# The argument is 'timeout and 'pod name' in seconds'
+function check_pod_ready_state () {
```
chaitanyaenr, line 49: Maybe log that the script is waiting for x seconds for the pod to transition to ready state? Thoughts?

chaitanyaenr (approved):
LGTM, nice work @amitsagtani97.

PR #146, opened 2021-03-23
Fix env.sh to accept parameters for NODE_COUNT and PODS_PER_NODE

kedark3:
# Description

I and @mffiedler were trying to run node-density tests in our CI. When we ran it in the CI we saw that it was defaulting to 1 node and 100 pods all the time. After much of debugging I saw it was changed recently in this commit https://github.com/cloud-bulldozer/e2e-benchmarking/pull/144/commits/3f306b639034372d3cae155114f3be9f8ba7959b 

We think it is a parameter we would certainly want to modify in our end of the environments. 

Fixes # (issue)

To address above issue, the fix is to source the values from env vars for these parameters, and then set to default to default values, if env vars are not set.

@rsevilla87 @mohit-sheth @chaitanyaenr  kindly review.

### DIFF
--- a/ci/env.sh
+++ b/ci/env.sh
@@ -11,8 +11,8 @@ export RUNS=1
 export JOB_ITERATIONS=3
 export CLEANUP=true
 export CLEANUP_WHEN_FINISH=true
-export NODE_COUNT=1
-export PODS_PER_NODE=100
+export NODE_COUNT=${NODE_COUNT:-1}
+export PODS_PER_NODE=${PODS_PER_NODE:-100}
 
 #for upgrade perf
 export VERSION=`oc get clusterversion | grep -o [0-9.]* | head -1`


mohit-sheth:
yeah, I think the values were kept keeping in mind that only our team will use them
LGTM

mohit-sheth:
also you might want to use https://github.com/cloud-bulldozer/scale-ci-pipeline/blob/master/jjb/dynamic/node-density.yml for CI runs
here you can pass your own values to the variables through a properties file https://github.com/cloud-bulldozer/scale-ci-pipeline/blob/master/properties-files/kubelet-density.properties

jtaleric:
> # Description
> I and @mffiedler were trying to run node-density tests in our CI. When we ran it in the CI we saw that it was defaulting to 1 node and 100 pods all the time. After much of debugging I saw it was changed recently in this commit [3f306b6](https://github.com/cloud-bulldozer/e2e-benchmarking/commit/3f306b639034372d3cae155114f3be9f8ba7959b)
> 
> We think it is a parameter we would certainly want to modify in our end of the environments.
> 
> Fixes # (issue)
> 
> To address above issue, the fix is to source the values from env vars for these parameters, and then set to default to default values, if env vars are not set.
> 
> @rsevilla87 @mohit-sheth @chaitanyaenr kindly review.

so `ci/env.sh` is for the actual CI of the code here, not to be used in a `CI` environment. Maybe there is some confusion there? I don't have an issue with this PR, but maybe the intended use was by mistake?

rsevilla87:
Indeed, this ci.sh file is for the CI tests of this repo. Those parameters are already configurable in the actual node-density scripts.

mffiedler:
> Indeed, this ci.sh file is for the CI tests of this repo. Those parameters are already configurable in the actual node-density scripts.

Thanks for the clarification.  Closing this PR in light of this.   We will work on our own driver script for our CI.

PR #148, opened 2021-03-24
Deprecate router-perf in favor of router-perf-v2

rsevilla87:
### Description

Always happy to deprecate things


mohit-sheth (approved):
LGTM

PR #152, opened 2021-03-26
Allow passing a custom elasticsearch backend url to use for log forwarding

dry923:
### Description
Added the ability for a user to pass an external elasticsearch url to use for a custom log forwarder backend.

### DIFF
--- a/workloads/logging/deploy_logging_stack.sh
+++ b/workloads/logging/deploy_logging_stack.sh
@@ -7,7 +7,7 @@ source env.sh
 
 # Logging format
 log() {
-  echo ${bold}$(date -u):  ${@}${normal}
+  echo -e "\033[1m$(date "+%d-%m-%YT%H:%M:%S") ${@}\033[0m"
 }
 
 # Check if oc client is installed
@@ -28,8 +28,13 @@ fi
 
 function install() {
   # create cluster logging and elasticsearch resources
-  log "Creating cluster logging and elastisearch resources"
-  envsubst < ./files/logging-stack.yml | oc create -f -
+  if [[ $CUSTOM_ES_URL != "" ]]; then
+    log "Creating cluster logging with custom elasticsearch backend"
+    envsubst < ./files/logging-stack_custom_es.yml | oc create -f -
+  else
+    log "Creating cluster logging and elastisearch resources"
+    envsubst < ./files/logging-stack.yml | oc create -f -
+  fi
 }
 
 wait_time=0
@@ -71,13 +76,15 @@ while [[ $( oc get daemonset.apps/fluentd -n openshift-logging -o=jsonpath='{.st
     exit 1
   fi
 done
-log "Logging stack including Elasticsearch, Fluend and Kibana are up"
+log "Logging stack is up"
 
-# Expose the elasticsearch service
-echo "Exposing the elasticsearch service by creating a route"
-oc extract secret/elasticsearch --to=/tmp/ --keys=admin-ca --confirm -n openshift-logging
-cp files/elasticsearch-route.yml /tmp/elasticsearch-route.yml
-cat /tmp/admin-ca | sed -e "s/^/      /" >> /tmp/elasticsearch-route.yml
-oc create -f /tmp/elasticsearch-route.yml -n openshift-logging
-routeES=`oc get route elasticsearch -n openshift-logging -o jsonpath={.spec.host}`
-echo "Elasticsearch is exposed at $routeES, bearer token is needed to access it"
+if [[ $CUSTOM_ES_URL == "" ]]; then
+  # Expose the elasticsearch service
+  log "Exposing the elasticsearch service by creating a route"
+  oc extract secret/elasticsearch --to=/tmp/ --keys=admin-ca --confirm -n openshift-logging
+  cp files/elasticsearch-route.yml /tmp/elasticsearch-route.yml
+  cat /tmp/admin-ca | sed -e "s/^/      /" >> /tmp/elasticsearch-route.yml
+  oc create -f /tmp/elasticsearch-route.yml -n openshift-logging
+  routeES=`oc get route elasticsearch -n openshift-logging -o jsonpath={.spec.host}`
+  log "Elasticsearch is exposed at $routeES, bearer token is needed to access it"
+fi
--- a/workloads/logging/README.md
+++ b/workloads/logging/README.md
@@ -17,6 +17,11 @@ Ensure to have `KUBECONFIG` set to the proper path to your desired cluster.
 Default: `4.6`
 Update channel for the Elasticsearch and Cluster logging operators.
 
+### CUSTOM_ES_URL
+Default: ""
+The external Elasticsearch url to direct logs to
+NOTE: If set, internal ElasticSearch will not be configured
+
 ### ES_NODE_COUNT
 Default: 3
 Number of Elasticsearch nodes.
--- a/workloads/logging/env.sh
+++ b/workloads/logging/env.sh
@@ -2,6 +2,7 @@
 
 # Vars and respective defaults
 export CHANNEL=${CHANNEL:=4.6}
+export CUSTOM_ES_URL=${CUSTOM_ES_URL:=""}
 export ES_NODE_COUNT=${ES_NODE_COUNT:=3}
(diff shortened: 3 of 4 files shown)


workloads/logging/deploy_logging_stack.sh
```diff
@@ -28,8 +28,13 @@ fi
 
 function install() {
   # create cluster logging and elasticsearch resources
-  log "Creating cluster logging and elastisearch resources"
-  envsubst < ./files/logging-stack.yml | oc create -f -
+  if [[ $DEPLOY_ES == "true" ]]; then
```
chaitanyaenr, line 31: This will skip the logging operator/fluentd setup as well, maybe modify the logging-stack template to selectively install Elasticsearch when DEPLOY_ES is set? Thoughts?

workloads/logging/deploy_logging_stack.sh
```diff
@@ -28,8 +28,13 @@ fi
...
+    log "Creating cluster logging and elastisearch resources"
+    envsubst < ./files/logging-stack.yml | oc create -f -
+  elif [[ $CUSTOM_ES_URL != "" ]]; then
```
chaitanyaenr, line 34: What's the use case for using custom ES?

dry923: If you want to point it to an off-cluster ES server you would need to do this. The simplest example being our CI which points to our dev ES instance for the log test. That way we don't need to eat up space on the cluster.

workloads/logging/deploy_logging_stack.sh
dry923: As it stands, if you don't set DEPLOY_ES to true or provide a custom es url then the script will pretty much just run the cleanup and then timeout waiting for the logger to come up. We could 
a) key off the custom url only and otherwise default to deploying; 
b) Allow it to setup a stub cluster logging instance with pretty much no configuration
c) Make the cleanup optional so it could verify that the cluster logging is "up"

I think option c would provide the most flexibility long term but am open to other options as well

workloads/logging/deploy_logging_stack.sh
chaitanyaenr: Got it, thanks.

workloads/logging/deploy_logging_stack.sh
chaitanyaenr: The intention of this variable is to have the ability to just setup the logging operator, fluentd stack when ES is not used as the backend and forward the logs to the custom ES right? This conditional will skip the entire logging stack install like you mentioned. We can tweak the spec to conditionally skip the elasticsearch setup when custom ES url is defined: https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/logging/files/logging-stack.yml#L82. Thoughts?

dry923: @chaitanyaenr I inverted the if on the install to simply be if the custom es url is defined use it otherwise do the install per normal. Does that resolve your concerns?

chaitanyaenr (approved):
LGTM

workloads/logging/deploy_logging_stack.sh
```diff
@@ -71,13 +76,15 @@ while [[ $( oc get daemonset.apps/fluentd -n openshift-logging -o=jsonpath='{.st
...
-echo "Exposing the elasticsearch service by creating a route"
-oc extract secret/elasticsearch --to=/tmp/ --keys=admin-ca --confirm -n openshift-logging
-cp files/elasticsearch-route.yml /tmp/elasticsearch-route.yml
-cat /tmp/admin-ca | sed -e "s/^/      /" >> /tmp/elasticsearch-route.yml
-oc create -f /tmp/elasticsearch-route.yml -n openshift-logging
-routeES=`oc get route elasticsearch -n openshift-logging -o jsonpath={.spec.host}`
-echo "Elasticsearch is exposed at $routeES, bearer token is needed to access it"
+if [[ $CUSTOM_ES_URL == "" ]]; then
+  # Expose the elasticsearch service
+  echo "Exposing the elasticsearch service by creating a route"
```
rsevilla87, line 83: s/echo/log/

workloads/logging/deploy_logging_stack.sh
```diff
@@ -71,13 +76,15 @@ while [[ $( oc get daemonset.apps/fluentd -n openshift-logging -o=jsonpath='{.st
...
+  oc extract secret/elasticsearch --to=/tmp/ --keys=admin-ca --confirm -n openshift-logging
+  cp files/elasticsearch-route.yml /tmp/elasticsearch-route.yml
+  cat /tmp/admin-ca | sed -e "s/^/      /" >> /tmp/elasticsearch-route.yml
+  oc create -f /tmp/elasticsearch-route.yml -n openshift-logging
+  routeES=`oc get route elasticsearch -n openshift-logging -o jsonpath={.spec.host}`
+  echo "Elasticsearch is exposed at $routeES, bearer token is needed to access it"
```
rsevilla87, line 89: s/echo/log/

rsevilla87:
Seems like with the current implementation are not initially available
```
export CUSTOM_ES_URL=blablabla
$ ./deploy_logging_stack.sh 
Wed 7 Apr 10:45:41 UTC 2021: Checking if oc client is installed
NAME      VERSION                             AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.8.0-0.nightly-2021-04-05-174735   True        False         17h     Cluster version is 4.8.0-0.nightly-2021-04-05-174735
Wed 7 Apr 10:45:44 UTC 2021: Deteting openshift-logging/openshift-operators-redhat namespaces if exists
Wed 7 Apr 10:45:51 UTC 2021: Installing the necessary objects for setting up elastic and logging operators and creating a cluster logging instance
Wed 7 Apr 10:45:51 UTC 2021: Creating cluster logging with custom elasticsearch backend
namespace/openshift-operators-redhat created
namespace/openshift-logging created
operatorgroup.operators.coreos.com/openshift-operators-redhat created
operatorgroup.operators.coreos.com/cluster-logging created
subscription.operators.coreos.com/cluster-logging created
unable to recognize "STDIN": no matches for kind "ClusterLogging" in version "logging.openshift.io/v1"
unable to recognize "STDIN": no matches for kind "ClusterLogForwarder" in version "logging.openshift.io/v1"
```

dry923:
> Seems like with the current implementation some CRDs are not initially available
> 
> ```
> $ export CUSTOM_ES_URL=blablabla
> $ ./deploy_logging_stack.sh 
> Wed 7 Apr 10:45:41 UTC 2021: Checking if oc client is installed
> NAME      VERSION                             AVAILABLE   PROGRESSING   SINCE   STATUS
> version   4.8.0-0.nightly-2021-04-05-174735   True        False         17h     Cluster version is 4.8.0-0.nightly-2021-04-05-174735
> Wed 7 Apr 10:45:44 UTC 2021: Deteting openshift-logging/openshift-operators-redhat namespaces if exists
> Wed 7 Apr 10:45:51 UTC 2021: Installing the necessary objects for setting up elastic and logging operators and creating a cluster logging instance
> Wed 7 Apr 10:45:51 UTC 2021: Creating cluster logging with custom elasticsearch backend
> namespace/openshift-operators-redhat created
> namespace/openshift-logging created
> operatorgroup.operators.coreos.com/openshift-operators-redhat created
> operatorgroup.operators.coreos.com/cluster-logging created
> subscription.operators.coreos.com/cluster-logging created
> unable to recognize "STDIN": no matches for kind "ClusterLogging" in version "logging.openshift.io/v1"
> unable to recognize "STDIN": no matches for kind "ClusterLogForwarder" in version "logging.openshift.io/v1"
> ```

@rsevilla87 Yeah I opened an issue about that here: https://github.com/cloud-bulldozer/e2e-benchmarking/issues/154

Effectively we're trying to apply the cr to quickly after apply the crd. If you re-run it now it will work without an issue. I was going to look into fixing that outside of this PR as it is pre-existing to this.

workloads/logging/deploy_logging_stack.sh
dry923: fixed. thanks

workloads/logging/deploy_logging_stack.sh
dry923: fixed. thanks

rsevilla87 (approved):
LGTM!

PR #153, opened 2021-03-26
Store file in custom directory for snappy in scale_ci_diagnosis script

amitsagtani97:
### Description
Updated Scale Ci diagnosis script to accept directory path for snappy.

File directory path is provided is set using SNAPPY_FILE_DIR env variable, which will be used in this script.


mohit-sheth (approved):
lgtm

PR #156, opened 2021-03-29
Test routes

rsevilla87:
### Description
I have observed a performance degradation that happens during the first iteration of the first test-scenario (also [happens in CI](https://ec2-52-39-240-23.us-west-2.compute.amazonaws.com:8443/job/e2e-benchmarking-ci/61/consoleFull)). I think the reason could be that the static content is not cached by nginx during this first iteration causing it's an extra delay to serve it. This commit adds a new action to the benchmark `test_routes`  to trigger a cURL to all routes before triggering any workload.

We shouldn't observe a big performance degradation of this iteration. (CI can help to verify this)

### DIFF
--- a/workloads/router-perf-v2/common.sh
+++ b/workloads/router-perf-v2/common.sh
@@ -113,7 +113,7 @@ gen_mb_config(){
     local port=443
   fi
   (echo "["
-  while read n r s p t w; do
+  while read host; do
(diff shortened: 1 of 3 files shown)


rsevilla87:
PTAL @mohit-sheth

mohit-sheth:
LGTM

PR #161, opened 2021-04-07
Add retry logic to curl

rsevilla87:
### Description

Hope it fixes issues like:
```
curl: (52) Empty reply from server
```


mohit-sheth (approved):
LGTM

PR #162, opened 2021-04-07
tolerances as a list instead of a string

mohit-sheth:
### Description
added 
```
 nargs=1,
 ``` 
 to the parser. This will take the tolerance number from cli as a list instead of string

Also, ran black on the code for readability


dry923:
Looks good, however I really am not a fan of the extra commas at the end of dictionaries, lists, etc from the formatter.

mohit-sheth:
no opinion but https://stackoverflow.com/questions/12087742/should-i-add-a-trailing-comma-after-the-last-argument-in-a-function-call summarizes the reason @dry923

Issue #163, opened 2021-04-09
Running in OpenShift Pod based Jenkins Agent


kedark3:
Hi @rsevilla87  in my team we are planning to run these in the jenkins agent that runs as a pod on OpenShift. I have not previously tried to run docker/podman within an openshift pod. Do you happen to know a way to do so? 
I believe this script was designed by keeping in mind that it would be ran from a jump host machine in the scale lab and not a container/pod. 

https://github.com/cloud-bulldozer/e2e-benchmarking/blob/af7eb1db8089d4a6507087810097648e4d3f36a5/workloads/router-perf-v2/common.sh#L54

Alternatively, I was thinking using kube-burner binary directly within the jenkins agent pod. Would you be able to accept that as a proposed change if that works for us. 

Thanks.

rsevilla87:
Fixed in #164

PR #164, opened 2021-04-09
Adding ability to run kube-burner without the container

kedark3:

### Description

Connected to issue #163 
When we run the router perf workload we run it within a pod on OpenShift. Hence instead of trying to sping up another container within there, it seemed as a better option to download kube-burner binary and use it directly. 

### Testing

I have tested this script locally on my machine as our PSI Jenkins is not available to test. Nonetheless, It should work just fine. 

cc @rsevilla87 @mffiedler @mohit-sheth @chaitanyaenr

### DIFF
--- a/workloads/router-perf-v2/common.sh
+++ b/workloads/router-perf-v2/common.sh
@@ -1,4 +1,5 @@
 ENGINE=${ENGINE:-podman}
+KUBE_BURNER_RELEASE_URL=${KUBE_BURNER_RELEASE_URL:-https://github.com/cloud-bulldozer/kube-burner/releases/download/v0.9.1/kube-burner-0.9.1-Linux-x86_64.tar.gz}
 INFRA_TEMPLATE=http-perf.yml.tmpl
 INFRA_CONFIG=http-perf.yml
 KUBE_BURNER_IMAGE=quay.io/cloud-bulldozer/kube-burner:latest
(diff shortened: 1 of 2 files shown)


workloads/router-perf-v2/common.sh
```diff
@@ -51,7 +52,15 @@ get_scenario(){
 deploy_infra(){
   log "Deploying benchmark infrastructure"
   envsubst < ${INFRA_TEMPLATE} > ${INFRA_CONFIG}
-  ${ENGINE} run --rm -v $(pwd)/templates:/templates:z -v ${KUBECONFIG}:/root/.kube/config:z -v $(pwd)/${INFRA_CONFIG}:/http-perf.yml:z ${KUBE_BURNER_IMAGE} init -c http-perf.yml --uuid=${UUID}
+  if [[ ${ENGINE} == "local" ]]; then
+    log "Downloading kube-burner binary"
+    curl ${KUBE_BURNER_RELEASE_URL}
+    log "Extracting kube-burner binary"
+    tar -xvzf kube-burner.tar.gz ; rm -f kube-burner.tar.gz
```
rsevilla87, line 59: What about replacing this lines by something like this?

```shell
curl -LsS ${KUBE_BURNER_RELEASE_URL} | tar xz
```

kedark3: Thanks Raul. I tried that and I get ```gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now``` 
But I added `-LsS` , updating commit.

kedark3: found a mistake. Fixed it. Your suggestion works.

rsevilla87 (approved):
LGTM!

PR #171, opened 2021-04-23
Add worker node selection for labeling workload nodes

kedark3:

### Description

During execution of Router ingress perf testing, we found that if we no nodes are labeled as `workload`, we run into failure.

### Fixes

It fixes above issue by selecting one worker node at random and label it as a `workload` node. It is not the ideal solution for all usecases, but in our case all our worker nodes are of same instance type and in that case it would make sense to label any node at random.

cc: @chaitanyaenr @mohit-sheth @mffiedler

### DIFF
--- a/workloads/router-perf-v2/common.sh
+++ b/workloads/router-perf-v2/common.sh
@@ -89,6 +89,11 @@ tune_liveness_probe(){
 
 tune_workload_node(){
   TUNED_SELECTOR=$(echo ${NODE_SELECTOR} | tr -d {:})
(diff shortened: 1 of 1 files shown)


mohit-sheth (approved):
LGTM

chaitanyaenr (approved):
LGTM

kedark3:
Note: I think the things are failing due to same error with touchstone_compare and csv_gen and ingress-performance failed before reaching the point in the code which I changed

mohit-sheth:
https://ec2-52-39-240-23.us-west-2.compute.amazonaws.com:8443/job/e2e-benchmarking-ci/99/consoleFull
seems to stop ingress-performance abruptly

merging for now

PR #177, opened 2021-04-30
Fix run comapre dot sh

kedark3:
### Description
Follow up to #171 as that PR added the labels, but too late for the run to succeed. `http-scale-client` deployment waits for the node with correct label and the run fails. 

### Fixes
This PR moves the labeling logic up in the chain so the deployment can succeed.

@mffiedler @mohit-sheth @chaitanyaenr


mohit-sheth:
lgtm

PR #181, opened 2021-05-12
initialized pod_count

amitsagtani97:
### Description
pod_count is not initialized at the beginning, hence there is wrong calculation of already running pods.


chaitanyaenr (approved):
LGTM

PR #183, opened 2021-05-12
remove ERROR lines from touchstone output

mohit-sheth:
### Description
Removes ERROR lines in the compare.yaml files so that csv_gen.py can parse the yaml file to create a csv


chaitanyaenr (approved):
LGTM

PR #186, opened 2021-06-03
replace JOB_ITERATIONS with TEST_JOB_ITERATIONS in kube-burner-crd.yaml to avoid conflicts

dry923:
### Description

The JOB_ITERATIONS variable in kube-burner is utilized twice. Once as an input from the user and again in the cr that is applied. This can cause confusion and conflicts when the user sets JOB_ITERATIONS in the env.sh file. This PR changes the variable used within the kube-burner scripts to TEST_JOB_ITERATIONS and applies that in the cr. This avoids reuse of the variable.

### DIFF
--- a/workloads/kube-burner/common.sh
+++ b/workloads/kube-burner/common.sh
@@ -113,7 +113,7 @@ label_nodes() {
   if [[ ${1} == "heavy" ]]; then
     total_pod_count=$((total_pod_count / 2))
   fi
-  export JOB_ITERATIONS=${total_pod_count}
+  export TEST_JOB_ITERATIONS=${total_pod_count}
   log "Labeling ${NODE_COUNT} worker nodes with node-density=enabled"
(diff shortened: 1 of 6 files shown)


Harshith-umesh:
hey @dry923 will we need to change JOB_ITERATIONS here for the ci to work?: https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/ci/env.sh

dry923:
> hey @dry923 will we need to change JOB_ITERATIONS here for the ci to work?: https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/ci/env.sh

@Harshith-umesh shouldn't have to. I changed the variable that should be effectively hidden to the end user so it shouldn't be an issue with any existing scripts.

jtaleric (approved):
LGTM

mohit-sheth (approved):
LGTM

PR #189, opened 2021-06-11
Fix scale perf

amitsagtani97:
- Remove the existing scale pod and benchmark before each run.
- Set the RUN to 2 for scale perf CI.

### DIFF
--- a/workloads/scale-perf/run_scale_fromgit.sh
+++ b/workloads/scale-perf/run_scale_fromgit.sh
@@ -16,8 +16,8 @@ for x in $(seq 1 $_runs); do
         exit 1
       fi
     fi
-  
-
+    
+    
     if [[ $x -eq 1 && $size -eq $_init_worker_count ]]
     then
       # Don't try to scale down on the first iteration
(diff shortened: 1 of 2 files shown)


workloads/scale-perf/run_scale_fromgit.sh
```diff
@@ -4,10 +4,9 @@ set -x
 source ./common.sh
 source ../../utils/common.sh
 
-oc -n my-ripsaw delete benchmark/scale --ignore-not-found --wait
-
 # Scale up/down $_runs times
 for x in $(seq 1 $_runs); do
+  oc -n my-ripsaw delete benchmark/scale --ignore-not-found --wait
```
chaitanyaenr, line 9: Maybe delete all existing benchmarks in the ns instead of just scale to avoid grabbing wrong fields? Thoughts?

amitsagtani97: my-ripsaw namespace is already deleted as a first step. Scale benchmark is the only one present.

amitsagtani97:
CI was terminated due to time limits of CI run :(

chaitanyaenr (approved):
LGTM

whitleykeith (approved):
lgtm

PR #192, opened 2021-06-15
Scale hotfix

whitleykeith:
### Description

Fixes issue with the scale benchmark where subsequent runs dont' work. Also setting the RUNS to default to 1


chaitanyaenr (approved):
LGTM

PR #195, opened 2021-06-18
Remote config variables should get the env var value

rsevilla87:
### Fixes

Remote config feature does not work properly w/o this fix, as it's values takes empty values regardless the env vars were previously set.


jtaleric (approved):
LGTM

PR #211, opened 2021-07-30
Baremetal update

jdowni000:
### Description
First attempt to have network-perf tests check infrastructure type for BareMetal. 
### Fixes
Once BareMetal is detected as the infrastructure type, separate code is used to assign client and server worker nodes for the benchmark tests. Currently, it only assigns the first two worker nodes that are returned. The current code fails to assign the env vars as it is written for aws.
Secondly, I had to update PyYAML version as currently it seems to use a deprecated version that `yaml.FullLoader` does not exist and will cause the code to fail.

### DIFF
--- a/workloads/network-perf/common.sh
+++ b/workloads/network-perf/common.sh
@@ -49,8 +49,10 @@ export_defaults() {
   #If using baremetal we use different query to find worker nodes
   if [ "$baremetal" != "No resources found in openshift-machine-api namespace." ]; then
   log "Colocating uperf pods for baremetal"
-  export server=$(oc get nodes --no-headers -l node-role.kubernetes.io/worker | awk '{print $1}')
-  export client=$(oc get nodes --no-headers -l node-role.kubernetes.io/worker | awk '{print $2}')
+  export server=$(oc get nodes --no-headers -l node-role.kubernetes.io/worker | awk '{print $1}')[0]
+  export client=$(oc get nodes --no-headers -l node-role.kubernetes.io/worker | awk '{print $1}')[1]
+  echo $server
+  echo $client
   
   else
     # If multi_az we use one node from the two first AZs
--- a/workloads/network-perf/common.sh
+++ b/workloads/network-perf/common.sh
@@ -224,12 +224,16 @@ assign_uuid() {
 }
 
 run_benchmark_comparison() {
+  log "Begining benchamrk comparison"
   ../../utils/touchstone-compare/run_compare.sh uperf ${baseline_uperf_uuid} ${compare_uperf_uuid} ${pairs}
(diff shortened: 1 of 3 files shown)


workloads/network-perf/common.sh
```diff
@@ -43,29 +43,39 @@ export_defaults() {
...
-      export server=$(oc get node -l node-role.kubernetes.io/worker,node-role.kubernetes.io/workload!="",node-role.kubernetes.io/infra!="",topology.kubernetes.io/zone=${zones[0]} -o jsonpath='{range .items[*]}{ .metadata.labels.kubernetes\.io/hostname}{"\n"}{end}' | head -n1)
-      export client=$(oc get node -l node-role.kubernetes.io/worker,node-role.kubernetes.io/workload!="",node-role.kubernetes.io/infra!="",topology.kubernetes.io/zone=${zones[1]} -o jsonpath='{range .items[*]}{ .metadata.labels.kubernetes\.io/hostname}{"\n"}{end}' | tail -n1)
-    else
-      log "At least 2 worker nodes placed in different topology zones are required"
-      exit 1
-    fi
-  # If multi_az is disabled we use the two first nodes from the first AZ
+  #If using baremetal we use different query to find worker nodes
+  if [[ "${baremetalCheck}" == '"BareMetal"' ]]; then
+  log "Colocating uperf pods for baremetal"
```
smalleni, line 51: Bash doesn't really require indentation sure, but mind indenting these lines inside the if block for constancy with other blocks in the code?

workloads/network-perf/common.sh
```diff
@@ -43,29 +43,39 @@ export_defaults() {
...
+  export server=$(oc get nodes --no-headers -l node-role.kubernetes.io/worker | awk 'NR==1{print $1}')
+  export client=$(oc get nodes --no-headers -l node-role.kubernetes.io/worker | awk 'NR==2{print $1}')
```
smalleni, line 53: Do you need a check to ensure atleast two nodes are present?

workloads/network-perf/common.sh
jdowni000: spaces included now

workloads/network-perf/common.sh
jdowni000: I will write new logic to check at least two are present before entering this part of the code

jdowni000: I have updated to check for at least two nodes before continuing to set client/server nodes. Application will terminate gracefully with log output that there are not 2 workers nodes available if it does not see at least 2 minimum.

smalleni:
Do you mind squashing your commits cleanly, please?

jdowni000:
> Do you mind squashing your commits cleanly, please?

PR #219, opened 2021-08-17
Fix threshold for network latency and throughput tolerance

kedark3:

### Description
The README said default should be 10 but script was setting it to 5. 

### Fixes
Fix the defaults to match what README says it should be in https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/network-perf/README.md


mohit-sheth:
looks good

jtaleric (commented):
I know this is already merged... but this was not hard set... so the caller could of just changed the value... Personally, 10% might make sense for Cloud (AWS, GCP, etc). Baremetal, not so much.

kedark3:
@jtaleric I agree caller should send it properly, but I thought updating it is good to have consistency with what Readme claims. I was assuming it will take 10 automatically as readme suggested.

PR #229, opened 2021-09-06
Rename VERSION parameter to prevent conflict with benchmark-operator…

rsevilla87:
…s VERSION

### Description

Benchmark-operator's makefile uses VERSION as parameter, the upgrade workload is breaking CI.


Harshith-umesh (approved):
LGTM!

PR #234, opened 2021-09-08
Pprof data collection for kube-burner workloads

Harshith-umesh:
### Description
This PR is for enabling and collection of pprof files(golang profiling information for kube-burner workloads) in e2e. 
If the env var PPROF_COLLECTION is set to true then pprof files are collected on a regular basis configured by the parameter pprofInterval which can be set by the env var PPROF_COLLECTION_INTERVAL, the collected pprof files are stored on the pods. To copy the files from the pods to a local directory(configured by the parameter pprofDirectory which by default is pprof-data) a background process will run every one minute to check if any new pprof files have arrived on the pod and will copy any if so incrementally. 
Finally the pprof-data folder will be tarred and shipped to snappy if enabled.
 
To enable pprof we had to make minor changes in ripsaw and kube-burner : 
https://github.com/cloud-bulldozer/benchmark-operator/pull/639 
https://github.com/cloud-bulldozer/kube-burner/pull/109


workloads/kube-burner/common.sh
```diff
@@ -16,6 +16,16 @@ log() {
   echo -e "\033[1m$(date "+%d-%m-%YT%H:%M:%S") ${@}\033[0m"
 }
 
+collect_pprof() {
+  sleep 50
```
amitsagtani97, line 20: why we need this sleep?

workloads/kube-burner/common.sh
```diff
@@ -16,6 +16,16 @@ log() {
...
+  while [ $(oc get benchmark -n benchmark-operator kube-burner-${1}-${UUID} -o jsonpath="{.status.complete}") == "false" ]; do
+    log "-----------------------checking for new pprof files--------------------------"
+    oc rsync -n benchmark-operator $(oc get pod -n benchmark-operator | grep "kube-burner" | grep Running | awk '{print $1}'):/tmp/pprof-data $PWD/
+    sleep 60
```
amitsagtani97, line 24: Move this to first step in this while loop and remove sleep 50 from top ?

workloads/kube-burner/common.sh
```diff
@@ -136,3 +153,19 @@ cleanup() {
   oc delete ns -l kube-burner-uuid=${UUID}
 }
 
+get_pprof_secrets() {
+export certkey=`oc get secret -n openshift-etcd | grep "etcd-serving-ip" | head -1 | awk '{print $1}'`
```
amitsagtani97, line 157: indentation missing

amitsagtani97 (changes requested):

workloads/kube-burner/common.sh
Harshith-umesh: to let kube-burner initialize and get started with the workload

workloads/kube-burner/common.sh
Harshith-umesh: nope , the while condition , that is the oc get benchmark will fail since the kube-burner workload hasn't initialized yet

amitsagtani97: Maybe use something like this, as done in 
`local timeout=$(date -d "+${POD_READY_TIMEOUT} seconds" +%s)
  until oc get benchmark -n my-ripsaw kube-burner-${1}-${UUID} --ignore-not-found -o jsonpath="{.status.state}" | grep -q Complete; do
    sleep 1
    if [[ $(date +%s) -gt ${timeout} ]]; then
      log "Timeout waiting for job to be created"
      exit 1
    fi`

instead of hard-coding the sleep time

Harshith-umesh: it's not possible to measure the time / check since that part is handled by the kube-burner binary, and it usually takes around 30sec for the kube-burner job to start till the pprof measurement begins

workloads/kube-burner/common.sh
```diff
@@ -16,6 +16,16 @@ log() {
   echo -e "\033[1m$(date "+%d-%m-%YT%H:%M:%S") ${@}\033[0m"
 }
 
+collect_pprof() {
+  sleep 50
+  while [ $(oc get benchmark -n benchmark-operator kube-burner-${1}-${UUID} -o jsonpath="{.status.complete}") == "false" ]; do
+    log "-----------------------checking for new pprof files--------------------------"
+    oc rsync -n benchmark-operator $(oc get pod -n benchmark-operator | grep "kube-burner" | grep Running | awk '{print $1}'):/tmp/pprof-data $PWD/
```
rsevilla87, line 23: You can optimize this expression with something like:
```suggestion
    oc rsync -n benchmark-operator $(oc get pod -n benchmark-operator -o name -l benchmark-uuid=${UUID}):/tmp/pprof-data $PWD/
```

workloads/kube-burner/common.sh
```diff
@@ -136,3 +153,19 @@ cleanup() {
   oc delete ns -l kube-burner-uuid=${UUID}
 }
 
+get_pprof_secrets() {
+ export certkey=`oc get secret -n openshift-etcd | grep "etcd-serving-ip" | head -1 | awk '{print $1}'`
+ echo `oc extract -n openshift-etcd secret/$certkey`
```
rsevilla87, line 158: Why the leading echo?

workloads/kube-burner/common.sh
```diff
@@ -136,3 +153,19 @@ cleanup() {
   oc delete ns -l kube-burner-uuid=${UUID}
 }
 
+get_pprof_secrets() {
+ export certkey=`oc get secret -n openshift-etcd | grep "etcd-serving-ip" | head -1 | awk '{print $1}'`
```
rsevilla87, line 157: You don't need to export this one
```suggestion
 local certkey=`oc get secret -n openshift-etcd | grep "etcd-serving-ip" | head -1 | awk '{print $1}'`
```

workloads/kube-burner/kube-burner-crd.yaml
```diff
@@ -38,6 +38,12 @@ spec:
         VERIFY_OBJECTS: ${VERIFY_OBJECTS}
         ERROR_ON_VERIFY: ${ERROR_ON_VERIFY}
         MAX_WAIT_TIMEOUT: ${MAX_WAIT_TIMEOUT}
+        PPROF_COLLECTION: ${PPROF_COLLECTION}
+        PPROF_COLLECTION_INTERVAL: ${PPROF_COLLECTION_INTERVAL}
+        BEARER_TOKEN: ${BEARER_TOKEN}
+        KEY: ${KEY}
```
rsevilla87, line 44: Can you rename this one to PRIVATE_KEY?, I personally always try to avoid name variables KEY/key just in case is a reserved name and is a bit ambiguous.

workloads/kube-burner/run_maxnamespaces_test_fromgit.sh
```diff
@@ -11,10 +11,31 @@ export REMOTE_METRIC_PROFILE=${REMOTE_METRIC_PROFILE:-https://raw.githubusercont
...
+ tar -zcvf pprof.tar.gz ./pprof-data
+ 
+ export workload=kube-burner-maxnamespaces
+
+ export snappy_path="$SNAPPY_USER_FOLDER/$runid$platform-$cluster_version-$network_type/$workload/$folder_date_time/"
+ generate_metadata > metadata.json  
+ ../../utils/snappy-move-results/run_snappy.sh pprof.tar.gz $snappy_path
+ ../../utils/snappy-move-results/run_snappy.sh metadata.json $snappy_path
+ store_on_elastic
+fi
```
rsevilla87, line 40: We can move all this logic to it's own function in common.sh and just leave something like below in the workload specific files

```shell
if [[ ${ENABLE_SNAPPY_BACKUP} == "true" ]] ; then
  snappy_backup kube-burner-maxnamespaces
fi
```

workloads/kube-burner/README.md
```diff
@@ -39,6 +39,8 @@ All scripts can be tweaked with the following environment variables:
 | **CLEANUP_WHEN_FINISH** | Delete workload's namespaces after running it | false |
 | **KUBE_BURNER_IMAGE** | Kube-burner container image | quay.io/cloud-bulldozer/kube-burner:v0.13 |
 | **LOG_LEVEL**        | Kube-burner log level | info |
+| **PPROF_COLLECTION** | Collect and store pprof data locally | false |
+| **PPROF_COLLECTION_INTERVAL** | Intervals for which pprof data will be collected | 5min |
```
rsevilla87, line 43:
```suggestion
| **PPROF_COLLECTION_INTERVAL** | Intervals for which pprof data will be collected | 5m | 
```

rsevilla87 (changes requested):
Nice work @Harshith-umesh, I have added some comments/suggestions!

workloads/kube-burner/run_maxnamespaces_test_fromgit.sh
Harshith-umesh: yes i've been thinking about this as well, will do that

workloads/kube-burner/common.sh
Harshith-umesh: echo makes sure that the cert and key are displayed on the same line , if only oc extract .. then it displays on 2 lines.

rsevilla87 (approved):
LGTM!

PR #241, opened 2021-09-13
Adjusting comparison to key on BASELINE_UUID being set

dry923:
### Description
Changing comparison to key on BASELINE_UUID being set instead of COMPARE_WITH_GOLD

### Fixes
If a user sets the large or small scale uuid to compare to and sets compate with gold to false (since they do not want to fetch the data from ES) the comparison was being skipped. Keying off BASELINE_UUID being set allows for either situation to work as well as the user not setting any comparison.


mohit-sheth (approved):
lgtm, thanks @dry923

PR #242, opened 2021-09-13
Adding missing env variable to env.sh and README

dry923:
### Description
The COMPARE env variable was being used in utils/touchstone_compare/run_compare.sh but was not in the router env.sh or README. If this is not set and the user wants to compare data without the GOLD UUID no comparison is produced.


mohit-sheth (approved):
lgtm

PR #243, opened 2021-09-14
add ROSA variables to scale script

morenod:
Add ROSA related vars to create benchmark file with ROSA related parametrs

### DIFF
--- a/workloads/scale-perf/README.md
+++ b/workloads/scale-perf/README.md
@@ -1,19 +1,20 @@
 # Scale Scripts
 
 The purpose of the scale scripts is scale Openshift Cluster to a given size.
-It will scale up and down X number of times and report the timings to 
+It will scale up and down X number of times and report the timings to
 Elasticsearch
 
 Running from CLI:
 
 ```sh
-$ ./run_<test-name>_fromgit.sh 
+$ ./run_<test-name>_fromgit.sh
 ```
 
 ## Environment variables
 
 ### OPERATOR_REPO
 Default: `https://github.com/cloud-bulldozer/benchmark-operator.git`  
+
 Benchmark-operator repo that you want to clone
 
 ### OPERATOR_BRANCH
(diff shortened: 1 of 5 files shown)


workloads/scale-perf/README.md
```diff
@@ -56,21 +56,37 @@ Default: `false`
 Enable/Disable the ability to compare two runs. If set to `true`, the next set of environment variables pertaining to the type of test are required
 
 ### BASELINE_CLOUD_NAME
-Default: ``
```
dry923, line 59: @morenod can we leave the default values or was there a reson you removed them?

workloads/scale-perf/run_scale_fromgit.sh
```diff
@@ -54,6 +69,7 @@ spec:
       serviceaccount: scaler
       poll_interval: $_poll_interval
       post_sleep: $_post_sleep
+$(_add_rosa_params)
```
rsevilla87, line 72: This is a bit tricky, I think you should consider using an if else logic to create the Benchmark CRDs

workloads/scale-perf/README.md
morenod: Empty '' is breaking the .md file generation. If default is none, there is no default

workloads/scale-perf/run_scale_fromgit.sh
morenod: It is the only way to add conditionals to a cat EOF structure without replicating all the content

workloads/scale-perf/README.md
morenod: restored empty defaults

workloads/scale-perf/run_scale_fromgit.sh
morenod: what do you think now?

workloads/scale-perf/README.md
```diff
@@ -53,24 +60,59 @@ Enable/Disable collection of metadata
...
+Default: ''
+
+Environment where cluster is deployed (Production, Staging)
+
+### ROSA_TOKEN
+Default: ''
+
+Token used to login on ROSA
+
+### AWS_ACCESS_KEY_ID & AWS_SECRET_ACCESS_KEY
```
dry923, line 106: Extremely minor nit but can you separate these 2 into separate items? I had to re-read it 3 times before i realized how you wanted the key and id lol.

workloads/scale-perf/run_scale_fromgit.sh
```diff
@@ -16,76 +17,50 @@ for x in $(seq 1 $_runs); do
         exit 1
       fi
     fi
-    
-    
+
+
```
dry923, line 21: Bonus empty space

workloads/scale-perf/README.md
morenod: done

workloads/scale-perf/run_scale_fromgit.sh
morenod: done

dry923 (approved):
LGTM! Ignoring the router failure since this does not touch that.

PR #244, opened 2021-09-14
Benchmark-comparison standarization

rsevilla87:
### Description

Effort to invoke benchmark-comparison using the same code from the different benchmarks. Apart from that, transitioning of the current csv_gen logic and take advantage of the touchstone capabilities to generate CSVs with tolerance reports, and then importing them to google spreadsheets with the help of a simple script.


utils/touchstone-compare/run_compare.sh
```diff
@@ -22,9 +22,9 @@ set -x
 if [[ ${COMPARE_WITH_GOLD} == "true" ]] || [[ ${COMPARE} == "true" ]]; then
```
dry923, line 22: You removed all documentation for the COMPARE or COMPARE_WITH_GOLD params but we still trigger on them.

rsevilla87: Indeed, I removed it from the router-perf-v2 workload, that workload does not use those parameters anymore. Network-performance still uses the "old" method.

rsevilla87: After a second read, I realized you were right, I didn't remove COMPARE_WITH_GOLD variables from the router-perf workload, removing them now.

rsevilla87:
Update: I have added a new variable **COMPARISON_RC** to handle benchmark-comparison return code in case some sample don't meet the tolerancy check.

mohit-sheth:
the CSV generated looks good and it is readable but creating graphs (which is the easiest way to convey our data outside our team) seems difficult.
Is there a way we can process the data further and create tables where we can generate graphs with one click?

Out current implementation has a lot of duplication (i.e uperf and router CSVs) but getting graphs out of them is a breeze

workloads/router-perf-v2/README.md
```diff
@@ -59,28 +59,30 @@ It's possible to tune the default configuration through environment variables. T
...
 | SERVICE_TYPE          | K8S service type to use | `NodePort` |
 | METADATA_COLLECTION   | Collect metadata prior to trigger the workload | `true` |
-| COMPARE_WITH_GOLD | Set it to true if the baseline uuids need to be from gold-index| "" |
-| ES_GOLD | Elasticsearch server where the gold index is stored | `ES_SERVER` |
-| GOLD_SDN | SDN that you want to compare your current run to | `openshiftsdn` |
-| GOLD_OCP_VERSION | OCP version for the gold baseline run | "" |
+
+### Benchmark comparison variables:
+
+The ingress-performance script is able to invoke benchmark-comparison to perform results comparisons and then generate a google spreadsheet document. If `ES_SERVER_BASELINE` is not set, benchmark-comparion is used to generate a CSV results file.
```
Harshith-umesh, line 68: typo- comparison

utils/touchstone-compare/run_compare.sh
Harshith-umesh: @rsevilla87 do we need this part in utils , since currently network perf does not use utils/touchstone-compare and when we will refactor network-perf we will use the standardized version, so is this required?

PR #225, opened 2021-08-31
Baremetal

jdowni000:
### Description
Baremetal workload.  PR for #208 . Has the ability to use new MCP's or not using supplied values or defaults. Calculates allocatable CPU per MCP and uses that number to deploy n number of sample-apps into each mcp. Than a mb-pod is created to hit all the replicas with http traffic for 1 days times (configurable). This will simulate http traffic and load in the cluster during an upgrade for testing.

PR #248, opened 2021-09-15
Baremetal e2e scripts

### LINKED ISSUE #208: Improve Update tests
The current upgrade tests at https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/upgrade-perf/run_upgrade_fromgit.sh seem to be too simplistic. I would like to see this improve based on the recent upgrade testing we had done for a customer. It is important to load up the cluster and here are some suggested improvements to the script.

1. Use kube-burner to load up the cluster with 100 projects using the cluster-density test
2. For each MachineConfig Pool that is not master, calculate the total allocatable CPUs for all the nodes in that MCP
3. Per MCP create a a new project and launch a sample HTTP node.js app through a deployment with enough replicas such that it consumes 50% of allocatable cores in the MCP 
4. Expose each app through a service
5. Expose the service as a route
6. Enforce network policy per app
7. Hit the routes for the duration for the upgrade test using mb[https://github.com/jmencak/mb]

Some yamls for sample 
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sampleapp
spec:
  replicas: 300
  selector:
    matchLabels:
      app: sample
  template:
    metadata:
      labels:
        app: sample
    spec:
      containers:
      - name: app
        image: quay.io/smalleni/sampleapp:latest
        readinessProbe:
          httpGet:
            path: /
            port: 8080
          initialDelaySeconds: 3
        ports:
        - containerPort: 8080
          protocol: TCP
        resources:
          requests:
            cpu: "1"
          limits:
            cpu: "1"
      nodeSelector:
        app: "true"
```

```
apiVersion: v1
kind: Service
metadata:
  name: samplesvc
spec:
  selector:
    app: sample
  ports:
  - port: 80
    targetPort: 8080
```

```
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: except
spec:
  podSelector:
    matchLabels:
      app: sample
  ingress:
  - from:
    - ipBlock:
        cidr: 10.128.0.0/14
        except:
        - "10.130.36.0/23"
        - "10.130.12.0/23"
        - "10.128.18.0/23"
        - "10.131.10.0/23"
        - "10.131.22.0/23"
        - "10.128.24.0/23"
        - "10.128.14.0/23"
```

mukrishn:
### Description
- Modifications to run multiple benchmark resources for baremetal cluster, so included conditional checks for operators.
- New workload for baremetal upgrade, that would bring up multiple application pods, services and active http traffic between them while upgrade. 

### Fixes
#208 

Developed by @jdowni000, I just fixed few things and corrected conflicts as he is on PTO. PR #225 can be marked as obsolete.


mohit-sheth:
rerun all

workloads/network-perf/common.sh
```diff
@@ -43,11 +44,41 @@ export_defaults() {
...
+  if [[ "${baremetalCheck}" == '"BareMetal"' ]]; then
+    log "BareMetal infastructure: setting isBareMetal accordingly"
+    export isBareMetal=true
+  else
+    export isBareMetal=false
+  fi
+
+  #If using baremetal we use different query to find worker nodes
+  if [[ "${isBareMetal}" == "true" ]]; then
+    log "Colocating uperf pods for baremetal"
```
smalleni, line 62: We are not really colocating are we?

workloads/kube-burner/common.sh
```diff
@@ -12,21 +17,34 @@ fi
...
-  oc delete namespace benchmark-operator --ignore-not-found
-  log "Cloning benchmark-operator from branch ${OPERATOR_BRANCH} of ${OPERATOR_REPO}"
-  rm -rf benchmark-operator
-  git clone --single-branch --branch ${OPERATOR_BRANCH} ${OPERATOR_REPO} --depth 1
-  (cd benchmark-operator && make deploy)
-  kubectl apply -f benchmark-operator/resources/backpack_role.yaml
-  kubectl apply -f benchmark-operator/resources/kube-burner-role.yml
-  log "Waiting for benchmark-operator to be running"
-  oc wait --for=condition=available "deployment/benchmark-controller-manager" -n benchmark-operator --timeout=300s
+  if [[ "${isBareMetal}" == "false" ]]; then
```
whitleykeith, line 33: tbh i think we should just leave the operator around each time and change this to apply to all cluster types. I don't see a reason why we need to delete/recreate the operator

workloads/network-perf/common.sh
```diff
@@ -224,31 +268,39 @@ assign_uuid() {
...
 }
 
 init_cleanup() {
-  log "Cloning benchmark-operator from branch ${operator_branch} of ${operator_repo}"
-  rm -rf /tmp/benchmark-operator
-  git clone --single-branch --branch ${operator_branch} ${operator_repo} /tmp/benchmark-operator --depth 1
-  oc delete -f /tmp/benchmark-operator/deploy
-  oc delete -f /tmp/benchmark-operator/resources/crds/ripsaw_v1alpha1_ripsaw_crd.yaml
-  oc delete -f /tmp/benchmark-operator/resources/operator.yaml  
+  if [[ "${isBareMetal}" == "false" ]]; then
```
whitleykeith, line 284: same comment as my other one. I think we should just default to leaving the operator around.

whitleykeith (changes requested):

workloads/network-perf/common.sh
mukrishn: right, we are not. we randomly pick 2 worker nodes. I will delete this log.

workloads/kube-burner/common.sh
```diff
@@ -12,21 +17,34 @@ fi
 export TOLERATIONS="[{key: role, value: workload, effect: NoSchedule}]"
 export UUID=$(uuidgen)
 
-log() {
-  echo -e "\033[1m$(date "+%d-%m-%YT%H:%M:%S") ${@}\033[0m"
-}
```
rsevilla87, line 17: Why this change? the original implementation should work in every platform with a recent version of bash. In addition, where are the bold and normal vars defined?

mukrishn: I believe this change was done to keep consistent datetime log across workload, understanding the original implementation was intentional so this will be removed.

workloads/kube-burner/common.sh
mukrishn: ++

rsevilla87: Once the operator is deployed, we shouldn't delete it, I'd modify the current code to check if operator is running, if so, do nothing, otherwise deploy it.

mukrishn: Removed delete command

mukrishn:
@mohit-sheth @rsevilla87 @whitleykeith any more suggestions/comments ?

whitleykeith (approved):
lgtm

PR #251, opened 2021-09-17
Sleep 10s before next run

rsevilla87:

### Description

We're creating a new benchmark with the same uuid just after deleting the previous one, benchmark-operator goes crazy and marks the new one as completed w/o doing nothing.


morenod (approved):
LGTM

morenod:
Passed scale test

PR #253, opened 2021-09-22
CPU and memory by node role plus failed pods breakdown

rsevilla87:

### Description
Add per node role memory/cpu metrics, also adding failed pod metrics to ease debugging.
Remove buffers + cached memory (usually useless)


mohit-sheth (approved):
lgtm

PR #256, opened 2021-09-29
making e2e scripts use manual uuid if given

whitleykeith:
Ensures e2e checks for the UUID env var before generating a random uuid. If that environment variable exists, it'll use that instead.


rsevilla87 (approved):
LGTM!

PR #257, opened 2021-09-30
Get kube-burner configuration from configmaps

rsevilla87:
### Description

The current implementation is not CI safe, as it doesn't allow to test a change. Thanks to this new feature recently added to kube-burner. We get the configuration files from configmaps that are rendered (envsubst) and created (kubectl create configmap) on the fly.

Also avoid removing the benchmark-operator upon creating the benchmark.


jtaleric (approved):
LGTM

workloads/kube-burner/workloads/node-pod-density/node-pod-density.yml
```diff
@@ -51,28 +50,25 @@ global:
...
-
-
 {{ end }}
-
 jobs:
   - name: node-density
-    jobIterations: {{.JOB_ITERATIONS}}
-    qps: {{.QPS}}
-    burst: {{.BURST}}
+    jobIterations: ${JOB_ITERATIONS}
```
Harshith-umesh, line 56: shouldn't this be TEST_JOB_ITERATIONS? node-pod-density.yml is used by both pod and node density light versions but node density uses JOB_ITERATIONS but pod density uses TEST_JOB_ITERATIONS, this might lead to confusion . can we just use either one ?

PR #260, opened 2021-10-07
add error handling to gsheet creation

amitsagtani97:
### Description
At present router and network test are marked as failed, if there are some issues with google sheet API's which should not be the case.
Recently came across a DAG run where EMAIL_ID var was not even exported but still it tried to send gsheet email to NULL address, which returned error and test was marked failed.


Harshith-umesh:
closing since this fix has been taken care by https://github.com/cloud-bulldozer/e2e-benchmarking/pull/244

PR #266, opened 2021-10-15
Rename metrics files

rsevilla87:

### Description

Airflow uses .yaml files rather than .yml


morenod (approved):
LGTM

PR #267, opened 2021-10-18
remove set

whitleykeith:
using `set -x` messes up the last output line of Airflow benchmarks, preventing the uuid from being passed and subsequently indexed. That breaks report functionality


Harshith-umesh:
Closing since we decided on making changes to the airflow script directly in a different PR.

PR #273, opened 2021-10-28
remove remaining ns after grace period

amitsagtani97:
### Description
force delete the remaining namespaces after 600s grace-period.


mohit-sheth:
looks like we need to delete the pods that are stuck in Terminating, this will then delete the ns @amitsagtani97

jtaleric:
yeah, this was not enough.

amitsagtani97:
@mohit-sheth @jtaleric 
this command 'oc delete --all pods -n $ns --force --grace-period=0 --ignore-not-found --wait;' 
wait until all pods(any state) are deleted and then deletes the ns in the next space.

PR #278, opened 2021-11-02
WIP Fixing the cleanup

jtaleric:
### Description
The current cleanup still left pods in the namespace, so it never would
cleanup properly.


dry923 (approved):
LGTM!

PR #281, opened 2021-11-10
Replace uuid variable that was empty with one with the actual UUID

dry923:
### Description
The benchmark_uuid was not in use anywhere and never set by the test so when it would go to compare the uuid of the running test would be empty. Replaced this with the UUID set for the run.


mohit-sheth (approved):
LGTM

Harshith-umesh:
closing since this fix is outdated , we no longer use the compare_uperf_uuid var or the assign uuid function.

PR #283, opened 2021-11-11
Adjust NUM_NODES command in router test to account for managed services

dry923:
### Description
The way NUM_NODES is calculated currently doesn't account for clusters that tag workload and infra nodes as worker nodes as well. This would result in a miscalculation of workers.

### Fixes
Fix NUM_NODES calculation to account for managed services


dry923:
no idea why the node tests failed given I only changed something in the router workload. re-running test.

chentex (approved):
LGTM

sjug (approved):

PR #287, opened 2021-11-18
fix relative path for google spreadsheet

Harshith-umesh:
### Fixes
wrong path for csv_gen.py

### DIFF
--- a/utils/common.sh
+++ b/utils/common.sh
@@ -78,7 +78,7 @@ gen_spreadsheet() {
   python -m venv ${csv_tmp}
   source ${csv_tmp}/bin/activate
   pip install oauth2client>=4.1.3 gspread
-  ./csv_gen.py --sheetname ${1}-$(date "+%Y-%m-%dT%H:%M:%S") -c ${2} --email ${3} --service-account ${4}
+  python3 ../../utils/csv_gen.py --sheetname ${1}-$(date "+%Y-%m-%dT%H:%M:%S") -c ${2} --email ${3} --service-account ${4}
   deactivate
   rm -rf ${csv_tmp}
 }


sjug (changes requested):
This is not the way

sjug:
Do we still want to keep the `csv_gen` functionality in e2e?

Harshith-umesh:
> Do we still want to keep the `csv_gen` functionality in e2e?

@sjug this is more of an immediate fix, cause router perf on airflow is failing.

mohit-sheth:
we want e2e just to handle pushing to the gsheet part. i.e no parsing of yaml or deciding pass/fail like we used to earlier. 
This PR https://github.com/cloud-bulldozer/benchmark-comparison/pull/55 should create a csv for us from benchmark-comparison

mohit-sheth (approved):
LGTM

Harshith-umesh:
> we want e2e just to handle pushing to the gsheet part. i.e no parsing of yaml or deciding pass/fail like we used to earlier. This PR [cloud-bulldozer/benchmark-comparison#55](https://github.com/cloud-bulldozer/benchmark-comparison/pull/55) should create a csv for us from benchmark-comparison

also gsheet functionality is not part of the workload code itself but handled in e2e-benchmarking/utils

morenod:
Im not 100% happy with this solution, but I need it to let the router tests to be completed

jtaleric (commented):
Huge -2 here. If I didn't have a doc appt, i would of blocked this totally.

jtaleric:
> a) Like I mentioned this isn't the correct way to fix this IMO. b) Removing the functionality would also be an immediate fix.

ALL OF THIS. 

This needs to be removed ASAP.

whitleykeith:
I tend to agree with @jtaleric and @sjug on this one too. 

IMO I think we should start looking at what we use the gsheet logic for and either (1) changing that logic to build visualizations from the data in ES (similar to the control plane dashboard we have for kube-burner results) or (2) adding the csv_gen logic into the report dag

mohit-sheth:
reverting this until all of us are on the same page

Issue #290, opened 2021-11-30
Router tests failing on deactivating virtualenv


morenod:
[2021-11-29, 22:56:37 EST] {subprocess.py:89} INFO - Google Spreadsheet link -> https://docs.google.com/spreadsheets/d/1whYNQ1tjYoQdYGGGSod2-O1XQbAO5EK7GhnIsmvICqw
[2021-11-29, 22:56:37 EST] {subprocess.py:89} INFO - [1mTue Nov 30 03:56:37 UTC 2021 Removing touchstone[0m
[2021-11-29, 22:56:37 EST] {subprocess.py:89} INFO - ../../utils/compare.sh: line 14: deactivate: command not found

Harshith-umesh:
fixed with https://github.com/cloud-bulldozer/e2e-benchmarking/pull/291

PR #291, opened 2021-11-30
fix venv deactivate conflict

Harshith-umesh:
### Description
error: `[2021-11-29, 20:38:02 UTC] {subprocess.py:89} INFO - ../../utils/compare.sh: line 14: deactivate: command not found`. 
This error occurs since we're trying to deactivate venv when it's already been deactivated. This happens because we're creating a virtualenv inside an already existing venv and when we try to deactivate the inner venv, both venv's get deactivated. The second venv is created in generate google spreadsheet function. 
first venv: https://github.com/cloud-bulldozer/e2e-benchmarking/blob/dc5b31b5119605579ccbd1c6eaf6bf4a2f81dc2c/utils/compare.sh#L8 
second venv: https://github.com/cloud-bulldozer/e2e-benchmarking/blob/dc5b31b5119605579ccbd1c6eaf6bf4a2f81dc2c/utils/common.sh#L78 
So a solution is to ignore the first venv's deactivate command if deactivate from the google spreadsheet fn is called. This can be done by checking if gsheet_key_location var is set , if not set we can let the deactivate command of the first venv execute.  
### Fixes
https://github.com/cloud-bulldozer/e2e-benchmarking/issues/290

### DIFF
--- a/utils/compare.sh
+++ b/utils/compare.sh
@@ -11,7 +11,9 @@ install_touchstone() {
 }
 
 remove_touchstone() {
-  deactivate
+  if [[ -z ${GSHEET_KEY_LOCATION} ]]; then
+    deactivate
+  fi
   rm -rf "${touchstone_tmp}"
 }
 
--- a/utils/compare.sh
+++ b/utils/compare.sh
@@ -11,9 +11,7 @@ install_touchstone() {
 }
 
 remove_touchstone() {
-  if [[ -z ${GSHEET_KEY_LOCATION} ]]; then
-    deactivate
-  fi
+  deactivate
   rm -rf "${touchstone_tmp}"
 }
 
--- a/utils/common.sh
+++ b/utils/common.sh
@@ -80,7 +80,7 @@ gen_spreadsheet_helper() {
 
 gen_spreadsheet() {
   log "Installing requirements to generate spreadsheet"
-  if [[ "$VIRTUAL_ENV" != "" ]]; then
+  if [[ "${VIRTUAL_ENV}" != "" ]]; then
     gen_spreadsheet_helper ${1} ${2} ${3} ${4}
   else
     csv_tmp=$(mktemp -d)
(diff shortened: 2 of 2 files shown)


morenod (approved):
LGTM

utils/compare.sh
```diff
@@ -11,7 +11,9 @@ install_touchstone() {
 }
 
 remove_touchstone() {
-  deactivate
+  if [[ -z ${GSHEET_KEY_LOCATION} ]]; then
```
sjug, line 14: Wouldn't it make more sense to check something like... `${VIRTUAL_ENV}`?

sjug:
> This happens because we're creating a virtualenv inside an already existing venv

Can we not just.. use one venv?

Harshith-umesh:
> > This happens because we're creating a virtualenv inside an already existing venv
> 
> Can we not just.. use one venv?

this gives me an idea, since gsheet fn is the last fn to execute before the program control flow goes to the remove touchstone fn which has the deactivate for the first venv, we can just remove the second venv deactivate from the gsheet fn. i tried this out and is feasible.

sjug:
While I like this solution better than the last, creating a second venv is unnecessary and is the root cause of the issue at hand.
Lets keep this and add some conditionals around the venv creation.

Harshith-umesh:
> While I like this solution better than the last, creating a second venv is unnecessary and is the root cause of the issue at hand. Lets keep this and add some conditionals around the venv creation.

I'm presuming the reasoning behind the 2nd venv was if they wanted to run that fn as a standalone fn, don't think we'll ever be doing that so makes sense to remove the 2nd venv completely.

sjug:
No need to break it, just check if we're already in a venv first.

Harshith-umesh:
> No need to break it, just check if we're already in a venv first.

this doesn't break anything. as you said we don't really need another venv. i guess this happened due to copy pasting this fn from an older code.

PR #296, opened 2021-12-07
Uperf single run

whitleykeith:
### Description

Current Uperf Scripts for pod and service network tests actually run three separate runs per script invocation (different number of pairs). In Airflow we assume 1 benchmark per task, so we want to run those three in separate tasks. 

This script `run.sh` will just run any generic uperf workload, and take in the necessary params to configure it. We can use this in our Airflow benchmark configs to split out the runs into separate tasks.

### DIFF
--- a/workloads/network-perf/run.sh
+++ b/workloads/network-perf/run.sh
@@ -3,11 +3,11 @@
 source ./common.sh
 export WORKLOAD=${WORKLOAD}
 
-  if [[ $WORKLOAD == "hostnet" ]]; then
+if [[ $WORKLOAD == "hostnet" ]]; then
     export HOSTNETWORK=true
-  elif [[ $WORKLOAD == "" ]]; then
+elif [[ $WORKLOAD == "service" ]]; then
     export SERVICEIP=true 
-  fi
+fi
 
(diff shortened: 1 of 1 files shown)


workloads/network-perf/run.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+source ./common.sh
+export WORKLOAD=${WORKLOAD}
+
+if [[ $WORKLOAD == "hostnet" ]]; then
+    export HOSTNETWORK=true
+elif [[ $WORKLOAD == "service" ]]; then
+    export SERVICEIP=true 
+fi
```
Harshith-umesh, line 10: else if $WORKLOAD == "pod" condition missing for pod-network-test

whitleykeith: what gets set from that conditional? the other workload vars changed another env but i don't see a var to change

Harshith-umesh: right

Harshith-umesh:
LGTM

mohit-sheth (approved):

PR #302, opened 2022-01-13
ACS: Disable metrics that take too long to index

gavin-stackrox:
### Description

These metrics take too long to index (>32mins) and this causes tasks to fail. This PR disables them until they can be usefully aggregated.


mohit-sheth (approved):
LGTM

whitleykeith (approved):
Lgtm

PR #304, opened 2022-01-14
Netobserv perf tests

memodi:
### Description
Adding scripts for performance testing for network-observability.


memodi:
/assign @mffiedler @kedark3 
could you help review this PR?

jtaleric (changes requested):
if am reading the README, this is literally a wrapper, of a wrapper, of a wrapper? Is that correct?

smalleni:
Can this be done as part of the existing networking e2e scripts by passing an additional flag or something to script?

jtaleric:
> Can this be done as part of the existing networking e2e scripts by passing an additional flag or something to script?

Better yet, just maintain this outside of end to end?

memodi:
> if am reading the README, this is literally a wrapper, of a wrapper, of a wrapper? Is that correct?

The intention wasn't to create a wrapper around networking e2e tests, but rather to **reuse** the existing code for networking e2e benchmarking tests and only adding changes required for network observability (netobserv) tests since both are tightly coupled.

memodi:
> Can this be done as part of the existing networking e2e scripts by passing an additional flag or something to script?

I had thought about it and this was probably simplest solution I could think of without causing lots of code changes to existing networking scripts and by keeping additional code isolated for netobserv. It's basically calling the networking e2e scripts as is, the only change that is currently for networking is to call run_pod_network_test_fromgit.sh script with custom uperf for netobserv.

kedark3:
@memodi your PR looks good to me, but since this is a community repo, we may want to get consensus on the code location and structure. If consensus is to maintain this outside of this repo, we can talk more about it on Slack. 

@jtaleric do we have an issue with adding network obsv tests in the repo? I think putting it in this repo makes is accessible to potentially any other teams or if there's a need for perfscale eng team(or other adjacent teams) then having it in this repo would make the most sense in terms of visibility, easy access. Putting it in OCP QE repos may not help in that regard.

Harshith-umesh:
Closing after discussing this with @memodi, as e2e is not the best place for this code.

PR #308, opened 2022-01-25
clear temp/ripsaw directory before running ci tests

amitsagtani97:
### Description
CI fails when we update any dependency library version in ripsaw because of the existing old installed ripsaw in the CI node.

Adding clean steps before running any ci test to make sure that we do not have any temp ripsaw files from previous runs.

### DIFF
--- a/ci/run_ci.sh
+++ b/ci/run_ci.sh
@@ -26,12 +26,10 @@ Workload                | Test                           | Result | Runtime  |
 ------------------------|--------------------------------|--------|----------|
 EOF
 
-# Clear the /tmp/ripsaw* directory to avoid pip version conflicts due to existing temp files.
-rm -rf /tmp/ripsaw*
-
 test_rc=0
 for test in ${test_list}; do            
-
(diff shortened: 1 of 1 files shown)


ci/run_ci.sh
```diff
@@ -26,6 +26,9 @@ Workload                | Test                           | Result | Runtime  |
 ------------------------|--------------------------------|--------|----------|
 EOF
 
+# Clear the /tmp/ripsaw* directory to avoid pip version conflicts due to existing temp files.
+rm -rf /tmp/ripsaw*
```
dry923, line 30: 2 Comments:

1) Since the install_cli function in utils/benchmark-operator.sh hard codes the path to /tmp/ripsaw-cli we should use the full path name and not wildcard.

2) Since we run the install_cli function EACH test we should probably move the rm to the top of the for loop below so we clear it each time.

amitsagtani97: @dry923 Fixed :+1:

dry923 (approved):
LGTM!

rsevilla87 (approved):
LGTM!

PR #317, opened 2022-02-10
Expose the number of samples as an option for network workloads

josecastillolema:
### Description
In bare metal environments the oscillation between test runs is less significant compared with shared cloud ones.
Exposing the number of samples as an option will allow to run faster tests in bare metal scenarios.


mohit-sheth (approved):

mohit-sheth:
Hi @jose, can we have the variable as a part of readme too

josecastillolema:
@mohit-sheth added some workload env vars to the README

rsevilla87 (approved):
LGTM!

PR #319, opened 2022-02-16
Add num_pairs bucket and remove max aggregation

rsevilla87:

### Description

Will improve spreadsheet generation.


dry923 (approved):
LGTM!

smalleni:
Why are we removing the max aggregation?

rsevilla87:
> Why are we removing the max aggregation?

We don't use it :)

PR #320, opened 2022-02-17
Have uperf tests only use a single UUID instead of 1 per pair

dry923:
### Description

Uperf testing was generating a UUID for each "Pair." This made viewing the data very cumbersome as you would have to track and stitch together multiple UUIDS. This PR simplifies it to have a single UUID for each execution of run_*test.sh

### DIFF
--- a/workloads/network-perf/run_serviceip_network_policy_test_fromgit.sh
+++ b/workloads/network-perf/run_serviceip_network_policy_test_fromgit.sh
@@ -9,15 +9,16 @@ if [[ "${isBareMetal}" == "true" ]]; then
   export METADATA_TARGETED=true
 fi
 
+export UUID=$(uuidgen)
 for pairs in 1 2 4; do
-  export UUID=$(uuidgen)
   export pairs=${pairs}
-  COMPARISON_OUTPUT="${PWD}/service-networkpolicy-${pairs}-pairs.csv"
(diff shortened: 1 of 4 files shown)


rsevilla87 (approved):
LGTM!

workloads/network-perf/run_serviceip_network_policy_test_fromgit.sh
```diff
@@ -9,15 +9,16 @@ if [[ "${isBareMetal}" == "true" ]]; then
   export METADATA_TARGETED=true
 fi
 
+export UUID=$(uuidgen)
 for pairs in 1 2 4; do
```
rsevilla87, line 13: I think we should add MOAR vars to the script. Missing a PAIRS vars :)

We can address that in a different PR though

dry923: Yeah I noticed that as well. +1 to in another PR

jtaleric:
Overall, looks OK. Not sure I think we should handle this in the e2e scripts vs `benchmark-operator` but, this is the simplest solution for sure.

dry923:
Closing this in lieu of https://github.com/cloud-bulldozer/e2e-benchmarking/pull/324 which incorporates the same functionality as well.

PR #325, opened 2022-02-24
Modifying the default restricted SCC is not required

smalleni:
The Job to create server/client use the benchmark-operator scc
that has priveleged access. Moreover this command fails on Managed
Services clusters since you cannot change an existing SCC there.


mohit-sheth:
oh is {"allowHostNetwork": true}' by default?

rsevilla87:
> oh is {"allowHostNetwork": true}' by default?

```
$ oc get scc privileged -o yaml | grep -i allowHostNetwork
allowHostNetwork: true
``

rsevilla87 (approved):
LGTM

PR #324, opened 2022-02-23
Unify networking benchmarks

rsevilla87:
### Description

Join networking benchmarks into a single script ./run.sh and improve README.md

This is a breaking change, we should push changes to airflow (or use a specific e2e-benchmarking commit/version) before merging it.

PR #326, opened 2022-02-24
export SERVICEIP within hostnetwork script

smalleni:
Accidentally setting SERVICEUP to true for hostnetwork
test sresults in pods not starting. Since there is no concept
of servceip for host network tests, makes sense to override that
in the script.

### DIFF
--- a/workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
+++ b/workloads/network-perf/run_hostnetwork_network_test_fromgit.sh
@@ -3,6 +3,7 @@
 source ./common.sh
 export WORKLOAD=hostnet
 export HOSTNETWORK=true
+export SERVICEIP=false
 export pairs=1
 export UUID=$(uuidgen)
 


Harshith-umesh:
SERVICEIP is set to false here by default: https://github.com/cloud-bulldozer/e2e-benchmarking/blob/061f0b6267a3b2f14c2bc038da116d3b9df7d731/workloads/network-perf/env.sh#L16

smalleni:
> SERVICEIP is set to false here by default:
> 
> https://github.com/cloud-bulldozer/e2e-benchmarking/blob/061f0b6267a3b2f14c2bc038da116d3b9df7d731/workloads/network-perf/env.sh#L16
> 
> and only in serviceip test scripts it's overridden to true

That is true, I guess I was trying to prevent a user from shooting themselves in the foot when they change the env.sh file for serviceip tests and then try to execute hostnetwork tests where the pods wont even be scheduled if SERVICEIP=true.

Harshith-umesh:
we can close this pr in lieu of https://github.com/cloud-bulldozer/e2e-benchmarking/pull/324

smalleni:
> we can close this pr in lieu of #324

SGTM

PR #327, opened 2022-03-04
Print pod info on error

rsevilla87:

### Description
We're currently blind when a benchmark fails. This change should shed some light to the error detail.

### DIFF
--- a/utils/benchmark-operator.sh
+++ b/utils/benchmark-operator.sh
@@ -59,8 +59,8 @@ run_benchmark() {
     for pod in $(kubectl -n benchmark-operator get pod -l benchmark-uuid=${UUID} -o custom-columns="name:.metadata.name" --no-headers); do
       pod_log=${tmp_dir}/${pod}.log
       log "Writing pod logs in ${pod_log}"
-      kubectl logs --prefix --tail=30 ${pod}
-      kubectl logs --prefix --tail=-1 ${pod} >> ${pod_log}
+      kubectl logs -n benchmark-operator --prefix --tail=30 ${pod}
+      kubectl logs -n benchmark-operator --prefix --tail=-1 ${pod} >> ${pod_log}
     done
     remove_cli
     exit 1
--- a/utils/benchmark-operator.sh
+++ b/utils/benchmark-operator.sh
@@ -53,11 +53,14 @@ remove_benchmark_operator() {
 run_benchmark() {
   source ${ripsaw_tmp}/bin/activate
   if ! ripsaw benchmark run -f ${1} -t ${2}; then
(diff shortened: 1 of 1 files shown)


utils/benchmark-operator.sh
```diff
@@ -52,8 +52,15 @@ remove_benchmark_operator() {
...
   source ${ripsaw_tmp}/bin/activate
-  set -e
-  ripsaw benchmark run -f ${1} -t ${2}
-  set +e
+  if ! ripsaw benchmark run -f ${1} -t ${2}; then
+    log "Benchmark failed, dumping workload more recent logs to stdout"
+    kubectl -n benchmark-operator get pod -l benchmark-uuid=${UUID}
+    for pod in $(kubectl -n benchmark-operator get pod -l benchmark-uuid=${UUID} -o custom-columns="name:.metadata.name" --no-headers); do
+      log "Pod ${pod}"
+      kubectl logs --prefix --since=30s ${pod}
```
dry923, line 60: Instead of just dumping what could be a good amount of logs to stdout what about a file named after each pod into a errorLog dir? that way its easier to grep through later and its not lost when the session is gone.

rsevilla87: IIRC, benchmarks are executed within an airlflow worker pod, which is ephemeral :cry

dry923: its not ephemeral when i run it outside of airflow :-D

mkarg75: I don't follow - kubectl logs is called outside the pod, so it'd write the log file to the host, right?

dry923: how about an option for either? just worried about loss to my tmux buffer lol

dry923: @mkarg75 airflow is running e2e from w/in a pod. When the test is done that pod goes poof. So storing the file locally there doesnt help.

mkarg75: ah, I was missing the link to airflow. thanks

rsevilla87: We can just print the latest 20 lines rather than the latest 30 secs. That should work fine for every terminal buffer

rsevilla87: @dry923, I added a new commit to dump pod logs to files in addition to the 30 latest lines to stdout

dry923: sounds good @rsevilla87 . we can roll w/ that for now and adjust later if needed. :+1:

dry923 (approved):
LGTM!

Harshith-umesh (approved):
LGTM

morenod (approved):
LGTM

PR #330, opened 2022-03-07
Only collect metadata once per run instead of for each pair

dry923:
### Description

We are currently running metadata collection for each pair which is unneeded (especially they all under the same UUID). This PR sets METADATA_COLLECTION to false for each run after the first pair.


workloads/network-perf/run.sh
```diff
@@ -36,11 +36,16 @@ if [[ ${SERVICEIP} == "true" && ${SERVICETYPE} == "metallb" ]]; then
...
 for pairs in ${PAIRS}; do
   export PAIRS=${pairs}
+  if [ $count -gt 0 ]; then
+    export METADATA_COLLECTION=false
+  fi
   if ! run_workload ${CR}; then
     exit 1
   fi
+  ((count++))
 done
```
rsevilla87, line 49: I think this way is easier (It doesn't need counter):

```suggestion
for pairs in ${PAIRS}; do
  export PAIRS=${pairs}
  if ! run_workload ${CR}; then
    exit 1
  fi
  export METADATA_COLLECTION=false
done
```

rsevilla87 (changes requested):
Good idea, some comments above

workloads/network-perf/run.sh
dry923: Good call. That's what I get for putting in a PR before coffee :-)

rsevilla87:
Seems it needs some rebase @dry923

rsevilla87 (approved):
lgtm

PR #331, opened 2022-03-07
Add new workload to capture cluster install timings from multiple runs

amitsagtani97:
th### Description
Script to install the OpenShift cluster n times and capture different install timings.

TODO - 
- [ ] README


dry923:
@amitsagtani97 2 questions.

1)  You mention timings in the description but I'm not seeing where you actually capture that data. Is that somewhere else I'm not seeing or has the coffee just not kicked in for me yet etc?

2) This test seems like it would benefit greatly from parallelization so we don't have to wait for a sequence of 5 installs, etc to happen. I realize bash is not the best place for that normally so this may not be the ultimately best place for this script to exist. On the other hand you could look into something like the parallel function to help and keep it bash.

rsevilla87:
Hey @amitsagtani97,  can close this one of favor of https://github.com/cloud-bulldozer/scale-ci-deploy/pull/184?

amitsagtani97:
@rsevilla87 sure, closed!

PR #332, opened 2022-03-07
Add cluster info to ES documents for router-perf-v2 workload

jaredoconnell:
### Description
This pull request adds cluster info to the elasticsearch documents published by the router-perf-v2 workload.
The purpose of this change is to simplify comparisons between tests done on the same cluster.

Here is the output I got the last time I ran it:
```
{
  "_index" : "router-test-results",
  "_type" : "_doc",
  "_id" : "kz2GZX8BwuFk1uEOZSnR",
  "_score" : 36.95297,
  "_source" : {
    "termination" : "http",
    "test_type" : "http",
    "uuid" : "b294f8ae-fbb9-4bab-9c10-a9de2a2465e3",
    "cluster.id" : "fd47491b-1b19-4338-a708-18d88a366d4f",
    "cluster.name" : "joconnel-kube-burner-t6876",
    "cluster.ocp_version" : "4.9.12",
    "cluster.kubernetes_version" : "1.22",
    "cluster.sdn" : "OpenShiftSDN",
    "cluster.platform" : "AWS",
    "cluster.type": "rosa",
    "requests_per_second" : 683,
    "avg_latency" : 1453,
    "latency_95pctl" : 2278,
    "latency_99pctl" : 5999,
    "host_network" : "true",
    "sample" : "1",
    "runtime" : 60,
    "routes" : 1,
    "conn_per_targetroute" : 1,
    "keepalive" : 0,
    "tls_reuse" : true,
    "number_of_routers" : "2",
    "200" : 40989,
    "timestamp" : "2022-03-07T17:55:20.331204"
  }
} 
```

### Fixes
[](https://issues.redhat.com/browse/PERFSCALE)[PERFSCALE-1579](https://issues.redhat.com/browse/PERFSCALE-1579)

PR #333, opened 2022-03-08
New document containing cluster information

mkarg75:
### Description

This PR adds the gen_metadata function in common.sh which will collect various important cluster information and send them in a separate document to ElasticSearch. 
The function has 3 arguments: The benchmark name, the start date and the end date, so that a call to it could be

```
run_workload() {
  log "Deploying benchmark"
  START_DATE=$(date +%s)
  local TMPCR=$(mktemp)
  envsubst < $1 > ${TMPCR}
  run_benchmark ${TMPCR} ${TEST_TIMEOUT}
  local rc=$?
  END_DATE=$(date +%s)
  log "Generating metadata"
  gen_metadata $WORKLOAD $START_DATE $END_DATE
  if [[ ${TEST_CLEANUP} == "true" ]]; then
    log "Cleaning up benchmark"
    kubectl delete -f ${TMPCR}
  fi
  return ${rc}
}
```


utils/common.sh
```diff
@@ -106,3 +106,83 @@ gen_spreadsheet() {
...
+
+gen_metadata() {
+        BENCHMARK=$1
+        START_DATE=$2
+        END_DATE=$3
+
+        # construct all the required information
+        local VERSION_INFO=$(oc version -o json)
+        local INFRA_INFO=$(oc get infrastructure.config.openshift.io cluster -o json)
+        local PLATFORM=$(echo ${INFRA_INFO} | jq -r .spec.platformSpec.type)
```
morenod, line 128: As I have mentioned on https://github.com/cloud-bulldozer/e2e-benchmarking/pull/332#pullrequestreview-902739288, platform type is not enough to differentiate AWS clusters from ROSA ones.

We need to check the tags included on platformStatus:

```
       "platformStatus": {
            "aws": {
                "region": "us-west-2",
                "resourceTags": [
                    {
                        "key": "red-hat-clustertype",
                        "value": "rosa"
                    },
                    {
                        "key": "red-hat-managed",
                        "value": "true"
                    }
                ]
            },
            "type": "AWS"
        }
    }
```

morenod (changes requested):
added some info to the platform parameter

workloads/cyclictest/library.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
rsevilla87, line 1: Do we need this file here?

mkarg75: removed

utils/common.sh
mkarg75: Checking if resourceTags[0].value is rosa now and adding it as CLUSTERTYPE to the document, let me know if that's enough

morenod: yep, now it is perfect

morenod (approved):
LGTM

rsevilla87 (approved):
LGTM!

PR #335, opened 2022-03-09
New metric templates for Hypershift

mukrishn:
### Description
Need a custom metric profile to query metrics for a hypershift model cluster. 
They run multiple prom instances individually, so using existing platform connector agent to fwd all metrics to observability thanos instance and querying it from there by kube-burner. 
This script install agents only on workload cluster(HostedCluster), so enabling it on the management cluster is a pre-req(not required if its deployed by [airflow](https://github.com/cloud-bulldozer/airflow-kubernetes/pull/134)).
```
                   │
                   │
Management Cluster │ Hosted Cluster
                   │
 ┌────────────┐    │  ┌────────────┐
 │ Prometheus │    │  │ Prometheus │
 └──────┬─────┘    │  └──────┬─────┘
        │          │         │
 ┌──────┴──────┐   │  ┌──────┴──────┐
 │grafana-agent│   │  │grafana-agent│
 └──────┬──────┘   │  └──────┬──────┘
        │write               │write
        │      ┌──────┐      │
        └──────►Thanos◄──────┘
               └───▲──┘
                   │
                   │query
            ┌──────┴──────┐
            │ Kube-burner │
            └──────┬──────┘
                   │
                   │index
           ┌───────▼───────┐
           │ Elasticsearch │
           └───────────────┘
```

### DIFF
--- a/workloads/kube-burner/common.sh
+++ b/workloads/kube-burner/common.sh
@@ -11,7 +11,11 @@ if [[ ${INDEXING} == "false" ]]; then
   export METADATA_COLLECTION=false
   unset PROM_URL
 else
-  export PROM_TOKEN=$(oc -n openshift-monitoring sa get-token prometheus-k8s)
+  if [[ ${HYPERSHIFT} == "false" ]]; then
+    export PROM_TOKEN=$(oc -n openshift-monitoring sa get-token prometheus-k8s)
+  else
+    export PROM_TOKEN=""
+  fi
 fi
 export TOLERATIONS="[{key: role, value: workload, effect: NoSchedule}]"
 export UUID=${UUID:-$(uuidgen)}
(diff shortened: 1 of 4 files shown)


workloads/kube-burner/common.sh
```diff
@@ -11,7 +11,12 @@ if [[ ${INDEXING} == "false" ]]; then
   export METADATA_COLLECTION=false
   unset PROM_URL
 else
-  export PROM_TOKEN=$(oc -n openshift-monitoring sa get-token prometheus-k8s)
+  if [[ ${HYPERSHIFT} == "false" ]]; then
+    export PROM_TOKEN=$(oc -n openshift-monitoring sa get-token prometheus-k8s)
+  else
+    export PROM_TOKEN="dummytokenforthanos"
```
dry923, line 17: Solid token content :-)

smalleni, line 17: Can we do a connection check with the token and fail if its not valid - trying to avoid the case where user doesn't change this.

mukrishn: > Can we do a connection check with the token and fail if its not valid - trying to avoid the case where user doesn't change this.

Thanos Querier doesn't have a password yet, so literally any string will work. 

> Also why can't the script install grafana-agent on the management cluster as well?

As management cluster needs a different `KUBECONFIG` file, I thought its kinda overdo to switch config and install it on other cluster, btw `airflow` does it by default on all cluster.
I am okay to add another optional input to pass management config file and install it on the fly, lmk.

rsevilla87 (approved):
LGTM!

PR #336, opened 2022-03-10
regex updated for Memory and CPU metrics

mukrishn:
### Description
To exclude `container="POD"` OR `container=""`


mukrishn:
@rsevilla87 small update in regex query, PTAL.

rsevilla87 (approved):
Always happy to prevent indexing garbage. LGTM!

PR #338, opened 2022-03-13
Unify kube burner

rsevilla87:
### Description

Similar to networking benchmarks, unify kube-burner workloads into a single run.sh script. @Harshith-umesh, I had to make some adaptations for CI tests to work with input variables.


morenod:
Associated change on airflow https://github.com/cloud-bulldozer/airflow-kubernetes/pull/137

morenod (approved):
LGTM

workloads/kube-burner/run.sh
```diff
@@ -0,0 +1,85 @@
...
+case ${WORKLOAD} in
+  cluster-density)
+    WORKLOAD_TEMPLATE=workloads/cluster-density/cluster-density.yml
+    METRICS_PROFILE=${METRICS_PROFILE:-metrics-profiles/metrics-aggregated.yaml}
+    export TEST_JOB_ITERATIONS=${JOB_ITERATIONS:-1000}
+  ;;
+  node-density)
+    WORKLOAD_TEMPLATE=workloads/node-pod-density/node-pod-density.yml
+    METRICS_PROFILE=${METRICS_PROFILE:-metrics-profiles/metrics.yaml}
+    NODE_COUNT=${NODE_COUNT:-4}
```
dry923, line 17: @rsevilla87 Not specific to this pr as it is the current case as well but why do we default to 4 nodes? Wouldn't it make more sense to default to the current worker count?

workloads/kube-burner/run.sh
```diff
@@ -0,0 +1,85 @@
...
+     exit 1
+  ;;
+esac
+
+log "###############################################"
+log "Workload: ${WORKLOAD}"
+log "Metrics profile: ${METRICS_PROFILE}"
+log "QPS: ${QPS}"
+log "Burst: ${BURST}"
+log "Job iterations: ${TEST_JOB_ITERATIONS}"
```
dry923, line 61: Job iterations is not pertinent to the node density tests. We could move this line into an else of the below if statement.

workloads/kube-burner/run.sh
```diff
@@ -0,0 +1,85 @@
...
+fi
+log "###############################################"
+if [[ ${PPROF_COLLECTION} == "true" ]] ; then
+  delete_pprof_secrets
+  delete_oldpprof_folder
+  get_pprof_secrets
+fi 
+run_workload kube-burner-crd.yaml
+rc=$?
+if [[ ${WORKLOAD} == node-density* ]]; then
```
dry923, line 74: It is possible that a run of node density had been run prior and failed for some external reason. Thus leaving the nodes labeled. What do you think about unlabeling the nodes again before we start the workload to ensure we don't have a incorrect testing environment?

workloads/kube-burner/run.sh
```diff
@@ -0,0 +1,85 @@
...
+  max-services)
+    WORKLOAD_TEMPLATE=workloads/max-services/max-services.yml
+    METRICS_PROFILE=${METRICS_PROFILE:-metrics-profiles/metrics-aggregated.yaml}
+    export TEST_JOB_ITERATIONS=${SERVICE_COUNT:-1000}
+  ;;
+  custom)
+  ;;
+  *)
+     log "Unkonwn workload ${WORKLOAD}, exiting"
+     exit 1
```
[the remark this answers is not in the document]
dry923: Disregard I missed the line above

workloads/kube-burner/run.sh
```diff
@@ -0,0 +1,85 @@
...
+  ;;
+esac
+
+log "###############################################"
+log "Workload: ${WORKLOAD}"
```
dry923, line 57: Should we also specify the workload template and alert template if they are provided/different than the norm?

rsevilla87: Well, alert template could make sense. Workload template is deduced from the workload :)

workloads/kube-burner/run.sh
rsevilla87: you're 100% right, I set it to 4 nodes due to historical reasons, transitioning it to worker node count

workloads/kube-burner/run.sh
rsevilla87: Adding alert template though

dry923: Right. My thinking was if it was a custom workload they would supply a template file which may be useful to denote in the output

Harshith-umesh (approved):
LGTM

workloads/kube-burner/README.md
```diff
@@ -68,7 +70,7 @@ Each iteration creates the following objects:
 
 The `node-density` and `node-density-heavy` workloads support the following environment variables:
 
-- **NODE_COUNT**: Number of worker nodes to deploy the pods on. During the workload nodes will be labeled with `node-density=true`. Defaults to 4.
+- **NODE_COUNT**: Number of worker nodes to deploy the pods on. During the workload nodes will be labeled with `node-density=true`. Defaults to the number of worker nodes across the cluster
```
dry923, line 73: might want to note that we do not count nodes that are also marked as infra and workload in addition to worker

rsevilla87: fixed :)

dry923 (approved):
LGTM!

PR #339, opened 2022-03-14
Fix cluster type detection being too specific

jaredoconnell:
### Description
This does a better job detecting the cluster type in router-perf-v2 so it works in all scenarios. The other one was too specific.

### Fixes


jaredoconnell:
Here is an example ES document:
```
{
    "termination": "http",
    "test_type": "http",
    "uuid": "7b0b7c57-cdfe-4654-8a29-3c8cc586451e",
    "cluster.id": "f8245f9a-b94a-4402-9872-fb687e9444f3",
    "cluster.name": "mrnd-ovn-qlcfm",
    "cluster.ocp_version": "4.10.0-rc.3",
    "cluster.kubernetes_version": "1.23",
    "cluster.sdn": "OVNKubernetes",
    "cluster.platform": "AWS",
    "cluster.type": "rosa",
    "requests_per_second": 992,
    "avg_latency": 2008,
    "latency_95pctl": 3005,
    "latency_99pctl": 5802,
    "host_network": "true",
    "sample": "1",
    "runtime": 60,
    "routes": 2,
    "conn_per_targetroute": 1,
    "keepalive": 0,
    "tls_reuse": true,
    "number_of_routers": "2",
    "200": 59561
}
```

That is actual output, and it does not change the specification compared to the existing code. It only makes it work by fixing a bug.

rsevilla87 (approved):
LGTM!

PR #343, opened 2022-03-16
Align default pods per node to new standard of 245

dry923:
### Description
Per https://github.com/cloud-bulldozer/airflow-kubernetes/pull/145 we are aligning to 245 for our node density tests by default.


rsevilla87 (approved):
LGTM

PR #345, opened 2022-03-17
Don't exit 1 on benchmark error

rsevilla87:

### Description

We shouldn't directly exit 1 on benchmark error. return 1 instead will let the parent script perform the benchmark cleanup before exiting.


dry923 (approved):
LGTM!

morenod (approved):
Yes please, always cleanup

LGTM

dry923:
The failures in the runs are from touchstone not any code changed by this PR. They are also "random" to when they fail and need to be looked at outside this PR. Merging this as the code is good.

PR #346, opened 2022-03-21
Index benchmark metadata

rsevilla87:

### Description

Index benchmark metadata leveraging the run_benchmark helper. Some benchmarks such as router-perf does not use that wrapper, we'll have to implement this functionallity separately.
Also renaming the start_date to the common timestamp field


mkarg75:
LGTM!

rsevilla87:
> hehe, that's why I asked you :D

Sure, I was going nuts cuz I couldn't find any document :D

utils/common.sh
```diff
@@ -177,15 +178,16 @@ local METADATA=$(cat << EOF
 "total_nodes":"${TOTAL_NODES}",
 "sdn_type":"${SDN_TYPE}",
 "benchmark":"${BENCHMARK}",
-"start_date":"${START_DATE}",
+"timestamp":"${START_DATE}",
```
Harshith-umesh, line 181: why are we changing this?

workloads/kube-burner/common.sh
```diff
@@ -142,6 +136,9 @@ check_running_benchmarks() {
 }
 
 cleanup() {
+  log "Cleaning up benchmark"
+  kubectl delete -f ${TMPCR}
+  kubectl delete configmap -n benchmark-operator kube-burner-cfg-${UUID}
   if ! oc delete ns -l kube-burner-uuid=${UUID} --grace-period=600 --timeout=30m; then
```
Harshith-umesh, line 142: if we're doing this , should we close this pr?

rsevilla87: yep

utils/common.sh
rsevilla87: We generally use timestamp as date field in our indexes and grafana is only able to handle a time field (Except for uperf which is uperf_ts, that's something we'll have to change to stantarize)

Harshith-umesh: ohh right yeah, this was added recently

Harshith-umesh:
LGTM

PR #348, opened 2022-03-21
Add resource requests for CNI workloads

smalleni:
Helps with better scheduling


rsevilla87 (approved):
LGTM!

PR #349, opened 2022-03-22
Add temporary taint removal until taint bug is resolved

dry923:
There is currently a bug [https://bugzilla.redhat.com/show_bug.cgi?id=2035005] which effects the placement of pods during our tests. This bug can cause a RreferNoSchedule taint to be on some nodes, thus greatly impacting statistics like pod ready times. This PR adds a temporary function that removes the taint from all nodes


rsevilla87 (approved):
LGTM!

smalleni (approved):

PR #350, opened 2022-03-23
[WIP] Cluster density thresholds

rsevilla87:
### Description

First approach to fire alerts in cluster-density tests leveraging the kube-burner alerting mechanism, so we can improve regression detection from CI.

At the moment it's using a 5m range in avg_over_time, the final range it's still TBD.


morenod (approved):
LGTM

workloads/kube-burner/alerts-profiles/cluster-density.yml
```diff
@@ -0,0 +1,43 @@
+# etcd
```
dry923, line 1: I agree in general that etdc should respond w/in these time frames but in azure the disks are slower to the point where the thresholds have been adjusted so etcd doesnt ping-pong around nodes. Not sure if we want to have some logic to deal w/ Azure

dry923: Per our slack convo. An Azure profile will be added in a separate PR.

dry923 (approved):
LGTM!

PR #351, opened 2022-03-24
Cluster-density refactor

rsevilla87:
### Description
The current cluster-density workload is outdated, and not very
representative of a real case scenario. It makes an extensive usage of
builds, the extensive usage of this object generates lot of delays in
the workload because of long building times and can also lead to timeouts.
Also, generating such amount of build objects is not very representative
of a real case scenario.

Replacing most of these builds by deployments and more services will generate a
similar CPU usage in the control plane and will help to generate a more
deterministic load in the cluster.

Results seen on a 120 worker nodes cluster using executing 1000 iterations:
image

image

Next one is really interesting, because it represents how service setup latency gets degraded over the duration of the test.
image

image


smalleni:
@rsevilla87 I generally like the idea but would believe we should start having > 1 endpoint per service. Thoughts?

smalleni:
I like it, finally > 1 endpoint/service/ Thoughts about multiple ports/service as that would increase the number of endpoints/service further?

rsevilla87:
> I like it, finally > 1 endpoint/service/ Thoughts about multiple ports/service as that would increase the number of endpoints/service further?

Keep in mind that each service is backing 2 pods, and five services are created in each iteration. In a 1k scenario would mean 5000 services backing 10000 pods, as shown in the capture above, the programming latency got increased ~x2 in this scenario, in a 4k scenario that increase would be at least x8

smalleni:
> > I like it, finally > 1 endpoint/service/ Thoughts about multiple ports/service as that would increase the number of endpoints/service further?
> 
> Keep in mind that each service is backing 2 pods, and five services are created in each iteration. In a 1k scenario would mean 5000 services backing 10000 pods, as shown in the capture above, the programming latency got increased ~x2 in this scenario, in a 4k scenario that increase would be at least x8 in OpenShiftSDN

Right, I'm just talking about changing the service template to back multiple ports instead of just one port per service.

rsevilla87:
> > > I like it, finally > 1 endpoint/service/ Thoughts about multiple ports/service as that would increase the number of endpoints/service further?
> > 
> > 
> > Keep in mind that each service is backing 2 pods, and five services are created in each iteration. In a 1k scenario would mean 5000 services backing 10000 pods, as shown in the capture above, the programming latency got increased ~x2 in this scenario, in a 4k scenario that increase would be at least x8 in OpenShiftSDN
> 
> Right, I'm just talking about changing the service template to back multiple ports instead of just one port per service.

Hey Sai, I get your idea, however I don't understand what you want to achieve by adding more complexity to the workload. We have to be able to execute 4K iterations of this workload in a 252 worker node scenario.

FYI, these are some of the metrics I got from a 4k test on a 252 node cluster:

image
image
image
image

rsevilla87:
@smalleni, I've been thinking of your suggestion and after some testing, I've added an extra port to each service/port, I think it's worth for our regular testing

smalleni:
> @smalleni, I've been thinking of your suggestion and after some testing, I've added an extra port to each service/port, I think it's worth for our regular testing

Thanks @rsevilla87. I mostly asked for this based on customer deployments and previous experience with you know who :)

workloads/kube-burner/README.md
```diff
@@ -58,15 +58,13 @@ The `cluster-density` workload supports the environment variable **JOB_ITERATION
...
-- 1 deployment with 2 pod replicas (sleep) mounting two secrets each. deployment-2pod
-- 2 deployments with 1 pod replicas (sleep) mounting two secrets. deployment-1pod
-- 3 services, one pointing to deployment-2pod, and other two pointing to deployment-1pod. (As a consequence, 3 endpoint objects pointing to 4 pods are created)
-- 3 route. 1 pointing to the service deployment-2pod and other two pointing to deployment-1pod
-- 10 secrets. 2 of them mounted by the previous deployments.
-- 10 configMaps. 2 of them mounted by the previous deployments.
+- 1 imagestream
+- 1 build
+- 5 deployments with pod 2 replicas (sleep) mounting 4 secrets, 4 configmaps and 1 downwardAPI volume each
+- 5 services, each ont pointing to the TCP/8080 port of one of the previous deployments
```
smalleni, line 64: Update with new port details as well.

smalleni (changes requested):

dry923 (approved):
LGTM!

PR #356, opened 2022-03-28
have different uuids for different pairs

mohit-sheth:
This commit will avoid 3 different runs to the same uuid (http://mohit-sheth-playground-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=4.11-aws-sdn-data-plane&task_id=pod-network&execution_date=2022-03-28T13%3A49%3A08.151285%2B00%3A00)


mohit-sheth:
we can now select different pairs in the dashboard so same uuid is not a problem anymore
Thanks @dry923

PR #357, opened 2022-03-30
Include workload field as metadata

rsevilla87:

### Description

Including the workload name as metadata field will be useful to perform comparisons. The value of this field, namely WORKLOAD, is already set by the different benchmarks of this repo.


morenod:
This is going to be very useful on uperf dashboards

mkarg75:
LGTM

PR #358, opened 2022-03-30
nginx pod cpu, mem resource requests

mukrishn:
### Description

- Adding pod `resource` req for router-perf nginx server pods for even pod distribution.
- Overriding `$PAIRS` for hostnet uperf workload.


rsevilla87 (approved):
LGTM!

PR #359, opened 2022-03-30
increase connect time for the routes to load up

mohit-sheth:
5seconds was not enough as we got 
```
[2022-03-29, 22:52:16 EDT] {subprocess.py:89} INFO - [1mWed Mar 30 02:52:16 UTC 2022 Testing all routes before triggering the workload[0m
[2022-03-29, 22:53:17 EDT] {subprocess.py:89} INFO - curl: (52) Empty reply from server
```


sjug:
If the connectivity test is timing out there are likely cluster issues and the router test would not add much value.

rsevilla87:
Closing this one @mohit-sheth ?

mohit-sheth:
looks like the router pods take more than 3 mins to come up, closing this PR

PR #360, opened 2022-03-31
Make kube-burner QPS for ingress tests consistent with other benchmarks

smalleni:
All other benchmarks use a QPS and BURST of 20.


mohit-sheth (approved):
lgtm

PR #361, opened 2022-04-01
Disable validation-webhook daemonset for managed clusters

smalleni:
When running the node density workloads on managed clusters, starting
with 4.10.5 on ROSA, we are seeing an error like "admission webhook
"regular-user-validation.managed.openshift.io" denied the request" when trying add a label to a
node directly by using oc label node. The recommended way to add labels
to nodes on managed ROSA clusters is by editing the machinepool.
However, the Default machinepool cannot be edited to add labels and we
see an error like "Labels cannot be updated on the Default machine pool".
The only way to add a label is by disabling the
validation-webhook daemonset and thereby admission cotnrol in the
openshift-validation-webhook project that only exists on managed services
clusters. We disable the daemonset by adding a fake nodeSelector before
labeling the nodes and remove thenodeSelector after unlabeling the nodes.
Adding a nodeSelector on top of the existing nodeAffinity means that both
the conditions needs to be met for a pod to be scheduled. Also by adding
the nodeSelector, the spec is not overwritten during reconcillation whereas
changes to nodeAffinity are being overwritten.

### DIFF
--- a/workloads/kube-burner-ocp-wrapper/run.sh
+++ b/workloads/kube-burner-ocp-wrapper/run.sh
@@ -14,10 +14,31 @@ EXTRA_FLAGS=${EXTRA_FLAGS:-}
 UUID=$(uuidgen)
 KUBE_DIR=${KUBE_DIR:-/tmp}
 
+check_managed_cluster() {
+   status=$(oc get infrastructure/cluster -o=jsonpath='{.status.platformStatus.*.resourceTags[0]}')
+   if [[ $status =~ managed ]]; then
+      echo "Detected a Managed Cluster"
+      managed=true
+   fi
+}
+
+remove_managed_webhook_validation() {
+   echo "Disabling validation-webhook for Managed cluster"
+   oc patch -n openshift-validation-webhook daemonset validation-webhook -p '{"spec": {"template": {"spec": {"nodeSelector": {"non-existing": "true"}}}}}'
+   
+}
+
+add_managed_webhook_validation() {
+    echo "Enabling validation-webhook for Managed cluster"
+    oc patch -n openshift-validation-webhook daemonset validation-webhook --type json -p '[{ "op": "remove", "path": "/spec/template/spec/nodeSelector" }]'
+}
+
 download_binary(){
   KUBE_BURNER_URL=https://github.com/cloud-bulldozer/kube-burner/releases/download/v${KUBE_BURNER_VERSION}/kube-burner-${KUBE_BURNER_VERSION}-Linux-x86_64.tar.gz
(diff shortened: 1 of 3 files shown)


rsevilla87:
> When running the node density workloads on managed clusters, starting with 4.10.5 on ROSA, we are seeing an error like "admission webhook "regular-user-validation.managed.openshift.io" denied the request" when trying add a label to a node directly by using oc label node. The recommended way to add labels to nodes on managed ROSA clusters is by editing the machinepool. However, the Default machinepool cannot be edited to add labels and we see an error like "Labels cannot be updated on the Default machine pool". The only way to add a label is by disabling the validation-webhook daemonset and thereby admission cotnrol in the openshift-validation-webhook project that only exists on managed services clusters. We disable the daemonset by adding a fake nodeSelector before labeling the nodes and remove thenodeSelector after unlabeling the nodes. Adding a nodeSelector on top of the existing nodeAffinity means that both the conditions needs to be met for a pod to be scheduled. Also by adding the nodeSelector, the spec is not overwritten during reconcillation whereas changes to nodeAffinity are being overwritten.
> 
> Signed-off-by: Sai Sindhur Malleni [smalleni@redhat.com](mailto:smalleni@redhat.com)
> ### Description
> ### Fixes

curious, why do we need to label a node?

smalleni:
> > When running the node density workloads on managed clusters, starting with 4.10.5 on ROSA, we are seeing an error like "admission webhook "regular-user-validation.managed.openshift.io" denied the request" when trying add a label to a node directly by using oc label node. The recommended way to add labels to nodes on managed ROSA clusters is by editing the machinepool. However, the Default machinepool cannot be edited to add labels and we see an error like "Labels cannot be updated on the Default machine pool". The only way to add a label is by disabling the validation-webhook daemonset and thereby admission cotnrol in the openshift-validation-webhook project that only exists on managed services clusters. We disable the daemonset by adding a fake nodeSelector before labeling the nodes and remove thenodeSelector after unlabeling the nodes. Adding a nodeSelector on top of the existing nodeAffinity means that both the conditions needs to be met for a pod to be scheduled. Also by adding the nodeSelector, the spec is not overwritten during reconcillation whereas changes to nodeAffinity are being overwritten.
> > Signed-off-by: Sai Sindhur Malleni [smalleni@redhat.com](mailto:smalleni@redhat.com)
> > ### Description
> > ### Fixes
> 
> curious, why do we need to label nodes?
> 
> I assume this is used for the old node-density implementation available in e2e-benchmarking, the implementation based on the kube-burner's OCP wrapper doesn't need to label nodes anymore.
> 
> Rather than keep fixing and updating the old implementation we should encourage users to use the new implementation

Oh, I didn't realize that was the case, thanks for surfacing that.

Issue #362, opened 2022-04-01
router-perf: deploy only the objects for the configured TERMINATIONS


qiliRedHat:
Current behavior: No matter what we configured for TERMINATIONS, all 4 namespaces' objects are created. https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/router-perf-v2/http-perf.yml.tmpl

Expected: If I set TERMINATIONS to http, only namespace of http and the objects under it are created. Same for other TERMINATIONS combinations.

Why: A customer case needs to evaluate the performance while pod number increases for http request. Because currently we installs all 4 namespaces, the pod number is 4 times of the pods actually visited by mb. Workaround can be using a customized http-perf.yml.tmpl with only http, but I think it's good to have this built in.

Please kindly review this enhancement request, if it is ok, I can open a pr to do it.

rsevilla87:
Hi @qiliRedHat , I was already aware of this issue, (however I didn't fix it yet as you noted here). This could be implemented with golang-templates included in kube-burner. I already have some WIP code to fix it, let me wrap it up it now.

PR #363, opened 2022-04-01
Template out infrastructure to match with TERMINATIONS

rsevilla87:

### Description

Also shifting to local engine by default (It's the fastest one and does not depends on podman anymore), and updating to a recent kube-burner version which includes the new template functions introduced at https://github.com/cloud-bulldozer/kube-burner/pull/135

- Fixes: https://github.com/cloud-bulldozer/e2e-benchmarking/issues/362

### DIFF
--- a/workloads/router-perf-v2/env.sh
+++ b/workloads/router-perf-v2/env.sh
@@ -10,8 +10,8 @@ export ES_INDEX=${ES_INDEX:-router-test-results}
 NUM_NODES=$(oc get node -l node-role.kubernetes.io/worker,node-role.kubernetes.io/workload!=,node-role.kubernetes.io/infra!= --no-headers | grep -cw Ready)
 LARGE_SCALE_THRESHOLD=${LARGE_SCALE_THRESHOLD:-24}
 METADATA_COLLECTION=${METADATA_COLLECTION:-true}
-ENGINE=${ENGINE:-podman}
-KUBE_BURNER_RELEASE_URL=${KUBE_BURNER_RELEASE_URL:-https://github.com/cloud-bulldozer/kube-burner/releases/download/v0.11/kube-burner-0.11-Linux-x86_64.tar.gz}
+ENGINE=${ENGINE:-local}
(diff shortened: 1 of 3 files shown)


workloads/router-perf-v2/env.sh
```diff
@@ -10,8 +10,8 @@ export ES_INDEX=${ES_INDEX:-router-test-results}
 NUM_NODES=$(oc get node -l node-role.kubernetes.io/worker,node-role.kubernetes.io/workload!=,node-role.kubernetes.io/infra!= --no-headers | grep -cw Ready)
 LARGE_SCALE_THRESHOLD=${LARGE_SCALE_THRESHOLD:-24}
 METADATA_COLLECTION=${METADATA_COLLECTION:-true}
-ENGINE=${ENGINE:-podman}
-KUBE_BURNER_RELEASE_URL=${KUBE_BURNER_RELEASE_URL:-https://github.com/cloud-bulldozer/kube-burner/releases/download/v0.11/kube-burner-0.11-Linux-x86_64.tar.gz}
+ENGINE=${ENGINE:-local}
+KUBE_BURNER_RELEASE_URL=${KUBE_BURNER_RELEASE_URL:-https://github.com/cloud-bulldozer/kube-burner/releases/download/v0.15.4/kube-burner-0.15.4-Linux-x86_64.tar.gz}
```
dry923, line 14: Instead of updating this every time we release do we want to just set it to latest via something like
```
$ curl -s "https://api.github.com/repos/cloud-bulldozer/kube-burner/releases/latest" | jq -r '.assets[] | .browser_download_url' | grep Linux-x86_64
```

rsevilla87: I personally prefer to stick with a fixed version, that way we can push changes to kube-burner safely

mukrishn (approved):

workloads/router-perf-v2/env.sh
dry923: ack

PR #364, opened 2022-04-01
Add ocm-api-load e2e benchmark

venkataanil:
1. Support for snappy yet to be added
2. Benchmark comparison yest to be added
3. As we are not testing the OCP cluster, we are avoiding
   a) metadata collection
   b) pprof collection


workloads/api_load/common.sh
```diff
@@ -0,0 +1,65 @@
...
+    log "BareMetal infastructure: setting isBareMetal accordingly"
+    isBareMetal=true
+  else
+    isBareMetal=false
+  fi
+
+  if [[ "${isBareMetal}" == "true" ]]; then
+    #Installing python3.8
+    sudo yum -y install python3.8
+  fi
```
smalleni, line 27: Do you really need this?

workloads/api_load/common.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+
+source env.sh
+source ../../utils/common.sh
+source ../../utils/benchmark-operator.sh
+
+openshift_login
```
smalleni, line 7: Do we need to login to openshift for this? I don't believe we are running this against an OCP cluster right?

workloads/api_load/common.sh
```diff
@@ -0,0 +1,65 @@
...
+
+  if [[ "${isBareMetal}" == "true" ]]; then
+    #Installing python3.8
+    sudo yum -y install python3.8
+  fi
+}
+
+deploy_operator() {
+  deploy_benchmark_operator ${OPERATOR_REPO} ${OPERATOR_BRANCH}
+  oc adm policy -n benchmark-operator add-scc-to-user privileged -z benchmark-operator
```
smalleni, line 32: Are you using this SCC somewhere?

smalleni (changes requested):

workloads/api_load/api-load-crd.yaml
```diff
@@ -0,0 +1,26 @@
...
+      duration: ${DURATION}
+      rate: ${RATE}
+      output_path: ${OUTPUT_PATH}
+      aws_access_key: ${AWS_ACCESS_KEY}
+      aws_access_secret: ${AWS_ACCESS_SECRET}
+      aws_account_id: ${AWS_ACCOUNT_ID}
+      cooldown: ${COOLDOWN}
+      sleep: ${SLEEP}
+      test_list: 
+       ${WORKLOAD}:
```
mukrishn, line 26: Excuse my ignorance, Is it just a string, without dict values ? 
how to run a list of workloads ?

smalleni: Looks like the CR lets you run only workload then? This seems like the operator implementation limitation

workloads/api_load/common.sh
smalleni: I mean will we ever run this script from baremetal? I'm OK leaving it around :)

workloads/api_load/api-load-crd.yaml
mukrishn: Probably not running complex scenario, but it iterates them in code https://github.com/cloud-bulldozer/benchmark-operator/blob/master/roles/api_load/templates/api_load.yml#L65.

workloads/api_load/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+Workloads can be tweaked with the following environment variables:
+
+
+| Variable         | Description                         | Default |
+|------------------|-------------------------------------|---------|
+| **OPERATOR_REPO**    | Benchmark-operator repo         | https://github.com/cloud-bulldozer/benchmark-operator.git      |
+| **OPERATOR_BRANCH**  | Benchmark-operator branch       | master  |
+| **ES_SERVER**        | Elasticsearch endpoint          | https://search-perfscale-dev-chmf5l4sh66lvxbnadi4bznl3a.us-west-2.es.amazonaws.com:443|
+| **ES_INDEX**         | Elasticsearch index             | ripsaw-kube-burner|
```
chentex, line 38: Here the index is `ripsaw-api-load`

workloads/api_load/api-load-crd.yaml
```diff
@@ -0,0 +1,26 @@
...
+apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
+kind: Benchmark
+metadata:
+  name: api-load-${WORKLOAD}-${UUID}
+  namespace: benchmark-operator
+spec:
+  uuid: ${UUID}
+  elasticsearch:
+    url: ${ES_SERVER}
+    index_name: "api-load-$(UUID}"
```
chentex, line 11: Here should be also the same index.

workloads/api_load/common.sh
```diff
@@ -0,0 +1,65 @@
...
+
+source env.sh
+source ../../utils/common.sh
+source ../../utils/benchmark-operator.sh
+
+openshift_login
+
+export_defaults() {
+  network_type=$(oc get network cluster -o jsonpath='{.status.networkType}' | tr '[:upper:]' '[:lower:]')
+  export UUID=$(uuidgen)
```
chentex, line 11: Normally when we run a test from this all `workloads` have the same UUID, se we can differentiate one run from another, it would be nice to be able to check if it has been defined. But maybe internally that `uuidgen` does that, have not gone to see yet.

workloads/api_load/README.md
```diff
@@ -0,0 +1,80 @@
...
+| **TEST_CLEANUP**        | Remove benchmark CR at the end | true |
+| **GATEWAY_URL**      | Gateway url to perform the test against       | "https://api.integration.openshift.com |
+| **OCM_TOKEN**| OCM Authorization token |  |
+| **AWS_ACCESS_KEY**    | AWS access key          |  |
+| **AWS_ACCESS_SECRET**              | AWS access secret                     |       |
+| **AWS_ACCOUNT_ID**            | AWS Account ID, is the 12-digit account number |       |
+| **RATE**| Rate of the attack. Format example 5/s | 10/s |
+| **DURATION**         | Duration of each individual run in minutes | 1 |
+| **OUTPUT_PATH** | Output directory for result and report files | /tmp/results |
+| **COOLDOWN**         | Cooldown time between tests in seconds | 10 |
```
chentex, line 49: Like I see it now, if we are going to run just one workload at a time, this can be removed.

chentex (commented):
Some comments on the current state, since it's still a WIP, looking good so far!

Some general things:
- Missing ramping functionality
- AWS region could be customizable also

workloads/api_load/api-load-crd.yaml
chentex: benchmark-operator let's you run multiple at a time

workloads/api_load/common.sh
venkataanil: I will remove this (for code simplicity) as out team may not be using baremetal OCP deployment.

workloads/api_load/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+**`SNAPPY_DATA_SERVER_URL`**
+Default: ''
+The Snappy data server url, where you want to move files.
+
+**`SNAPPY_DATA_SERVER_USERNAME`**
+Default: ''
+Username for the Snappy data-server.
+
+**`SNAPPY_DATA_SERVER_PASSWORD`**
```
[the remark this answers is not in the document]
venkataanil: sure

workloads/api_load/common.sh
venkataanil: no. I will remove that.

workloads/api_load/api-load-crd.yaml
venkataanil: Each test (like this example config https://github.com/cloud-bulldozer/ocm-api-load/blob/main/config.example.yaml#L28 ) can have different values for "rate" and "duration" options. 
Benchmark-operator also allows this configuration https://github.com/cloud-bulldozer/benchmark-operator/blob/master/roles/api_load/templates/api_load.yml#L71
As e2e accepts only environment variables, I need to define a new env variable for each test's rate and duration options, for example,
ACCESS_TOKEN_RATE: 1000/h
ACCESS_TOKEN_DURATION: 1
LIST_SUBSCRIPTIONS_RATE: 2000/h
LIST_SUBSCIPTION_DURATION: 1
...
CERTIFICATES_RATE: 2/s
CERTIFICATES_DURATION: 1

Please let me know If we have any other way to get these values from user.

workloads/api_load/README.md
venkataanil: oh, typo :)

workloads/api_load/README.md
venkataanil: I will change it after we conclude to go with one or multiple tests by each e2e run.

workloads/api_load/api-load-crd.yaml
venkataanil: thanks Vicente.

workloads/api_load/api-load-crd.yaml
venkataanil: Currently we need to pass many environment variables to ./run.sh for invoking the test i.e
WORKLOAD=list-clusters GATEWAY_URL="http://localhost:8080" OCM_TOKEN="notARealToken" RATE=10/s AWS_ACCESS_KEY="empty" AWS_ACCESS_SECRET="empty" AWS_ACCOUNT_ID="empty" ./run.sh

instead I am thinking user can directly pass the config yaml file (i.e CONFIG=~/config.yaml ./run.sh) which run.sh parses and retrieve these variables. Another advantage is - yaml config file can define multiple tests with args
( https://github.com/cloud-bulldozer/ocm-api-load/blob/main/config.example.yaml#L31 )
 helping us run multiple tests with single invocation of run.sh

mukrishn: you could also set all these variables in `env.sh` file one time and just `./run.sh`  right.
And we already dealing such scenarios in some e2e scripts and [airflow](https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/tasks/benchmarks/e2e.py) takes care of setting those secrets/vars during the task execution. IMO passing env var is fine.

chentex: And `env.sh` file doesn't sound bad, probably tricky to create the first time, but once it's there, it would be just replacing stuff.

venkataanil:
> 
ramping and aws region are not available in benchmark-operator api load
https://github.com/cloud-bulldozer/benchmark-operator/blob/master/roles/api_load/templates/api_load.yml

workloads/api_load/api-load-crd.yaml
venkataanil: Please check latest approach

mukrishn (approved):
LGTM

chentex:
> > 
> 
> ramping and aws region are not available in benchmark-operator api load https://github.com/cloud-bulldozer/benchmark-operator/blob/master/roles/api_load/templates/api_load.yml

This PR takes care on adding those to benchmark operator, thank you.
https://github.com/cloud-bulldozer/benchmark-operator/pull/750

chentex (commented):
I just have a comment regarding the default values of each tests specific configuration.

workloads/api-load/env.sh
```diff
@@ -0,0 +1,54 @@
...
+export QUOTA_COST_RATE=${QUOTA_COST_RATE:-$RATE}
+export QUOTA_COST_DURATION=${QUOTA_COST_DURATION:-$DURATION}
+export RESOURCE_REVIEW_RATE=${RESOURCE_REVIEW_RATE:-$RATE}
+export RESOURCE_REVIEW_DURATION=${RESOURCE_REVIEW_DURATION:-$DURATION}
+export CLUSTER_AUTHORIZATIONS_RATE=${CLUSTER_AUTHORIZATIONS_RATE:-$RATE}
+export CLUSTER_AUTHORIZATIONS_DURATION=${CLUSTER_AUTHORIZATIONS_DURATION:-$DURATION}
+export SELF_TERMS_RATE=${SELF_TERMS_RATE:-$RATE}
+export SELF_TERMS_DURATION=${SELF_TERMS_DURATION:-$DURATION}
+export CERTIFICATES_RATE=${CERTIFICATES_RATE:-$RATE}
+export CERTIFICATES_DURATION=${CERTIFICATES_DURATION:-$DURATION}
```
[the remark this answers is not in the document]
venkataanil: Thanks. When I changed my code like this

    for attr in "${!p[@]}"; do
      if [[ ! -z "${p[$attr]}" ]]; then
        echo "          $attr: ${p[$attr]}"
      fi
    done

I am able to generate CRD file as you suggested here. I will update the PR now.

smalleni:
@chentex Is this good to go now?

chentex (approved):
LGTM

workloads/api-load/run.sh
```diff
@@ -0,0 +1,14 @@
...
+source ./common.sh
+
+CR=api-load-crd.yaml
+
+log "###############################################"
+log "api-load tests: ${TESTS}"
+log "###############################################"
+
+run_workload ${CR}
+
```
dry923, line 13: @venkataanil Can we add a removal of the benchmark operator at the end here? Something like this from the kube-burner one
```
remove_benchmark_operator ${OPERATOR_REPO} ${OPERATOR_BRANCH}
```

venkataanil: Thanks Russ. Agree, I will update the PR with your suggestion.

venkataanil: @dry923 I have gone through other workloads now and none of them are removing benchmark operator after the test is run. So lets follow the same methodology and skip calling remove_benchmark_operator.

venkataanil:
@dry923 can we merge this PR? I will revisit this later for the config file approach.

PR #366, opened 2022-04-06
script to install and create hosted cluster 

mukrishn:
### Description
standalone script to install hypershift operator on management cluster and create hostedclusters
Only for AWS based management cluster,


mukrishn:
@dry923 @morenod PTAL

workloads/hyper-scale/README.md
```diff
@@ -0,0 +1,31 @@
+# Hypershift perfscale Scripts
+
+The purpose of the script is to install Hypershift Operator on an Openshift cluster and create HostedCluster workload on them.
+
+1. Install and create cluster: `./run.sh build`
```
dry923, line 5: It might be a convenient option to also have this install the base hypershift cluster. Doesn't necessarily need to be in this PR but as a thought for the future.

workloads/hyper-scale/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Workload variables
+
+The run.sh script can be tweaked with the following environment variables
+
+| Variable                | Description              | Default |
+|-------------------------|--------------------------|---------|
+| **AWS_REGION**            | AWS region where the management cluster lives on | us-west-2 |
+| **AWS_ACCESS_KEY_ID**       | AWS access key for authentication | **REQUIRED** |
+| **AWS_SECRET_ACCESS_KEY**     | AWS secret for authentication | **REQUIRED** |
+| **ROSA_ENVIRONMENT**           | ROSA environment, staging or production        | staging |
```
dry923, line 24: At the moment it doesn't look like we actually need any of the rosa bits in here since we are not actually using the binary for anything productive.

workloads/hyper-scale/common.sh
```diff
@@ -0,0 +1,98 @@
...
+    while [[ $cm != "oidc-storage-provider-s3-config" ]]
+    do
+        cm=$(oc get configmap -n kube-public oidc-storage-provider-s3-config --no-headers | awk '{print$1}' || true)
+        echo "Hypershift Operator is not ready yet..Retrying after few seconds"
+        sleep 5
+    done
+}
+
+create_cluster(){
+    echo $PULL_SECRET > pull-secret
```
dry923, line 54: We might want to add an option to set a version/image for the hosted cluster.

workloads/hyper-scale/common.sh
```diff
@@ -0,0 +1,98 @@
...
+
+create_empty_cluster(){
+    echo $PULL_SECRET > pull-secret
+    hypershift create cluster none --name $HOSTED_CLUSTER_NAME --node-pool-replicas=0 --base-domain $BASEDOMAIN --pull-secret pull-secret --control-plane-availability-policy $REPLICA_TYPE --network-type $NETWORK_TYPE
+    echo "Wait till hosted cluster got created and in progress.."
+    oc wait --for=condition=available=false --timeout=60s hostedcluster -n clusters $HOSTED_CLUSTER_NAME
+    oc get hostedcluster -n clusters $HOSTED_CLUSTER_NAME
+}
+
+postinstall(){
```
dry923, line 69: With our tests we generally need to make an adjustment to firewall rules for certain network tests to be able to run post-install. We should look at add those, potentially here, as well. Example from airflow rosa install https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/scripts/utils/rosa_post_install.py

workloads/hyper-scale/run.sh
```diff
@@ -0,0 +1,49 @@
...
+prep
+if [[ "$1" == "clean" ]]; then
+    printf "Running Cleanup Steps"
+    setup
+    cleanup
+elif [[ "$1" == "build" ]]; then
+    setup
+    install
+
+    export NODEPOOL_SIZE=$COMPUTE_WORKERS_NUMBER
```
dry923, line 21: We should probably ensure that the worker count is divisible by 3 if HA is enabled

workloads/hyper-scale/README.md
mukrishn: you mean a script to install a base OSD(AWS) management cluster ? I am not sure if this repo is the right place to keep cluster install scripts, I can write up one and keep it in a separate repository. 
I treat this script as a hypershift workload script to spin/clean up `HostedControlPlane` and `HostedCluster` resources on an existing management cluster.

workloads/hyper-scale/common.sh
mukrishn: yes, I have an issue created for that on airflow repo, https://github.com/cloud-bulldozer/airflow-kubernetes/issues/160. 
Once that is ready, I will include here as well in a separate PR.

workloads/hyper-scale/run.sh
mukrishn: On Hypershift model, on HA `HostedControlPlane` pods are create in x3 as pods on the management cluster, so workers need not to be that way. I believe once @smalleni had a hostedcluster with 1 or 2 workers, isn't it ?

workloads/hyper-scale/README.md
dry923: Right. This wasn't a comment to the AWS account info but the actual rosa bits which aren't used at this point. We can leave them in there as a place holder but we should provide a comment around them saying such

workloads/hyper-scale/common.sh
dry923: sounds good

workloads/hyper-scale/run.sh
mukrishn: I also confirmed with engineers https://coreos.slack.com/archives/C01C8502FMM/p1649344956868999, `HostedCluster` worker nodes need not to be like that. At least not yet :smile:

mukrishn:
@morenod @dry923 any more comments or thoughts ?

dry923 (approved):
LGTM!

PR #368, opened 2022-04-06
Resetting cleanup when finish to false

paigerube14:
### Description
The default value of cleanup when finished was set to true by accident when merged. Resetting to false to reflect the table in the README


Harshith-umesh (approved):
LGTM

Issue #371, opened 2022-04-07
No module named yaml if failure in run_benchmark


paigerube14:
There is a module not found error being thrown in our scale-ci jobs if a failure occurs. Think it is because it has already removed the cli (deactivating the virtualenv) that installs the pyyaml requirement. There is a chance that the python being used does not have pyyaml installed 

Line here (I think) is causing error https://github.com/cloud-bulldozer/e2e-benchmarking/blob/33f90351dff67f9cea79c17aaa112e4e705e0e50/utils/benchmark-operator.sh#L70

```
04-07 11:40:14.299  [1mThu Apr  7 15:40:02 UTC 2022 Benchmark failed, dumping workload more recent logs[0m
04-07 11:40:14.299  NAME                         READY   STATUS   RESTARTS   AGE
04-07 11:40:14.299  kube-burner-1762b359-wdskw   0/1     Error    0          11m
........
Exiting kube-burner"
04-07 11:40:14.299  Traceback (most recent call last):
04-07 11:40:14.299    File "<string>", line 1, in <module>
04-07 11:40:14.299  ModuleNotFoundError: No module named 'yaml'
```

paigerube14:
No longer seeing this issue with PR 372

PR #372, opened 2022-04-07
Moving remove cli till after gen metadata and only if failure

paigerube14:
### Description
Moving remove_cli to after gen metadata so the python virtualenv is still avaiable with pyyaml package. 

Also adding if statement if project_name does not exist in the build helper (was seeing the error under fixes)

### Fixes
https://github.com/cloud-bulldozer/e2e-benchmarking/issues/371

```
04-07 11:40:36.224  No resources found in --no-headers namespace.
04-07 11:40:36.224  build_helper.sh: line 10: ${#project_name}-1: substring expression < 0
```

### DIFF
--- a/utils/benchmark-operator.sh
+++ b/utils/benchmark-operator.sh
@@ -65,7 +65,6 @@ run_benchmark() {
       kubectl logs -n benchmark-operator --prefix --tail=30 ${pod}
       kubectl logs -n benchmark-operator --prefix --tail=-1 ${pod} >> ${pod_log}
     done
-    remove_cli
   fi
   local benchmark_name=$(cat ${1} | python -c 'import yaml; import sys; print(yaml.safe_load(sys.stdin.read())["metadata"]["name"])')
   gen_metadata ${benchmark_name} ${start_date} $(date +%s%3N)
--- a/workloads/kube-burner/build_helper.sh
+++ b/workloads/kube-burner/build_helper.sh
@@ -3,20 +3,22 @@
 function prepare_builds_file()
 {
   project_name=`oc get project --no-headers | grep -m 1 conc-b | awk {'print $1'}`
-  bc_name=`oc get bc -n $project_name --no-headers | awk {'print $1'}`
-  running_build_file="running-builds.json"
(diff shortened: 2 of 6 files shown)


utils/benchmark-operator.sh
```diff
@@ -65,9 +65,12 @@ run_benchmark() {
       kubectl logs -n benchmark-operator --prefix --tail=30 ${pod}
       kubectl logs -n benchmark-operator --prefix --tail=-1 ${pod} >> ${pod_log}
     done
-    remove_cli
   fi
   local benchmark_name=$(cat ${1} | python -c 'import yaml; import sys; print(yaml.safe_load(sys.stdin.read())["metadata"]["name"])')
   gen_metadata ${benchmark_name} ${start_date} $(date +%s%3N)
+
+  if [[ -${rc} -eq 1 ]]; then
```
rsevilla87, line 72: Why did you add this conditional?

paigerube14: because the remove_cli was originally only in the if statement starting on line 57, so I was trying to keep the same functionality by putting an if statement

paigerube14:
BTW I think the second issue I listed under fixes is actually fixed by not running the incorrect workload ([PR 373](https://github.com/cloud-bulldozer/e2e-benchmarking/pull/373))

But still think the extra check for project name is helpful in the build_helper.py file

utils/benchmark-operator.sh
paigerube14: @rsevilla87 do you think we could move the _remove_cli_ to the end of the run.sh scripts in case other places need items installed by the benchmark operator requirements? There might also be other cases similar to the one I hit if we remove the cli but need a package that was installed in the virtualenv

Do we need to be calling remove_benchmark_operator instead?

paigerube14: Specifically for the concurrent-builds it runs _run_benchmark_ multiple times so I would have to add in a call to re-activate the cli

rsevilla87 (approved):
lgtm!

PR #373, opened 2022-04-07
Incorrect string comparison for workload type of run

paigerube14:
### Description
The string comparison of the type of workload was not working properly. All workloads were getting into the if statement and causing a bunch of problems with running workloads multiple times 

### Fixes
All workloads that aren't concurrent builds to run once and properly


rsevilla87 (approved):
Good catch, lgtm

PR #374, opened 2022-04-11
Set targeted to false for metadata collection by default

dry923:
### Description
Targeted metadata collection may run into issues with ports in use and fail. Setting this to targeted: false by default.


mohit-sheth (approved):
lgtm

mukrishn (approved):

PR #378, opened 2022-04-13
HAProxy: Add multiple backends per service to test balancing algorithms and their performance impact

sjug:
### Description
Addition of a new environment variable that allows us to adjust the number of replicas per back-end that would allow testing of the balancing algorithms.

### Fixes
Adds new (common) customer config for router performance testing.


sjug:
A few parameters need to be set for this new case.

DEPLOYMENT_REPLICAS (desired replicas per deployment), INFRA_TEMPLATE (http-perf-depl.yml.tmpl) and SMALL_SCALE_ROUTES or LARGE_SCALE_ROUTES as desired.

The *SCALE_ROUTES parameters will be the number of deployments.

workloads/router-perf-v2/env.sh
```diff
@@ -17,6 +17,9 @@ KUBE_BURNER_IMAGE=quay.io/cloud-bulldozer/kube-burner:latest
 TERMINATIONS=${TERMINATIONS:-"http edge passthrough reencrypt mix"}
 INFRA_TEMPLATE=http-perf.yml.tmpl
```
gcs278, line 18: It's my understand we have to change this ENV variables `INFRA_TEMPLATE`, but it always gets overridden right here no? Should you have:
`export INFRA_TEMPLATE=${INFRA_TEMPLATE:-http-perf.yml.tmpl}`

Same may go with `INFRA_CONFIG` if you think so.

sjug: We set the environment variables in `env.sh` (this file), which is sourced on every execution of `ingress-performance.sh`.
`INFRA_CONFIG` never really needs to change, that's just the output filename after the vars are parsed.

gcs278: Got it. I was pointing out that it doesn't follow the pattern of most every other variable that it uses, e.g. `${VARIABLE:-"default"}`. I don't modify env.sh with my test runs, I use a .envrc source file per test configuration, which all other variables allow me to override.

I'm guessing this was previously hardcoded because it was never intended to change, but now, we have a reason to change `INFRA_TEMPLATE`. Is there a reason you don't want `INFRA_TEMPLATE` to follow this pattern? Keep it like this just makes it less configurable.

workloads/router-perf-v2/env.sh
```diff
@@ -17,6 +17,9 @@ KUBE_BURNER_IMAGE=quay.io/cloud-bulldozer/kube-burner:latest
...
 INFRA_CONFIG=http-perf.yml
+export DEPLOYMENT_REPLICAS=${DEPLOYMENT_REPLICAS:-10}
+#export SMALL_SCALE_ROUTES=10
```
gcs278, line 21: Now I get why you put this hear, but it's kinda cryptic and I just glanced over it and thought nothing of it. Consider putting a comment such as:
`# Recommend to use a smaller number of routes if running with INFRA_TEMPLATE=http-perf-depl.yml.tmpl`

workloads/router-perf-v2/env.sh
sjug: You're right, there's no downside to changing it to match. I'm considering removing the `INFRA_CONFIG` var as well, as it is entirely irrelevant.

workloads/router-perf-v2/env.sh
sjug: Would the README be a better place for this?

gcs278: oh snap there's a README!? I'll let you decide on that, I'm good as long as it's somewhere I suppose. It could go in the SMALL_SCALE_ROUTES and LARGE_SCALE_ROUTES rows. 

Now that you mention there is a README, should you add `DEPLOYMENT_REPLICAS` as a new ENV variable and a explanation?

sjug: Indeed, I have added `DEPLOYMENT_REPLICAS` to the README (second commit to this PR).

gcs278: Ah sorry I see that now. Looks good.

PR #379, opened 2022-04-18
update deployment template with requests

mukrishn:
### Description
Deployment templates are not right, missing `resources` key for cpu mem requests.
`deployment.spec.template.spec.containers.request` >> `deployment.spec.template.spec.containers.resources.request`


rsevilla87 (approved):
good catch!

PR #380, opened 2022-04-19
Use unauthenticated repository for postgres image

rsevilla87:

### Description
We should use the unauthenticated repository to prevent auth issues


smalleni (approved):
LGTM

afcollins:
`registry.access.redhat.com/rhscl/postgresql-10-rhel7:latest` doesn't have an ARM variant :(

PR #384, opened 2022-04-26
Append UUID to result csv

smalleni:
This helps differentiate results in the case where
we do multiple runs of the router test from the same directory.


rsevilla87 (approved):
Makes sense, lgtm!

Issue #386, opened 2022-04-28
Empty string in Platform in AWS


rsevilla87:
The expression

https://github.com/cloud-bulldozer/e2e-benchmarking/blob/044c0b0b815b48c8e2f2bcb643dfe36f79830d05/utils/common.sh#L126-L131

Set platform as a empty string when cluster is AWS and not ROSA.

```
# echo ${INFRA_INFO} | jq .
{
  "apiVersion": "config.openshift.io/v1",
  "kind": "Infrastructure",
  "metadata": {
    "creationTimestamp": "2022-04-28T10:16:20Z",
    "generation": 1,
    "name": "cluster",
    "resourceVersion": "589",
    "uid": "24f457bb-f295-4fc6-b071-ae95b1831cd3"
  },
  "spec": {
    "cloudConfig": {
      "name": ""
    },
    "platformSpec": {
      "aws": {},
      "type": "AWS"
    }
  },
  "status": {
    "apiServerInternalURI": "lol",
    "apiServerURL": "lol",
    "controlPlaneTopology": "HighlyAvailable",
    "etcdDiscoveryDomain": "",
    "infrastructureName": "ci-4-11-aws-sdn-contr-qg6cr",
    "infrastructureTopology": "HighlyAvailable",
    "platform": "AWS",
    "platformStatus": {
      "aws": {
        "region": "us-west-2"
      },
      "type": "AWS"
    }
  }
}

# CLUSTERTYPE=$(echo ${INFRA_INFO} | jq -r '.status.platformStatus.aws.resourceTags | map(select(.key == "red-hat-clustertype"))[0].value' | tr '[:lower:]' '[:upper:]')

``

PR #387, opened 2022-04-28
Fix clustertype detection on AWS clusters

morenod:
When no resourceTags are defined, jq cannot iterate over null. Adding a try will create an empty result.

PLATFORM will be the result of the query to the resourceTags if result is not empty


morenod:
Fixes #386

Issue #388, opened 2022-04-28
Error rendering configuration template: "TERMINATIONS"


sjug:
After this PR landed: https://github.com/cloud-bulldozer/e2e-benchmarking/pull/363

I'm now seeing this error:
```
Thu Apr 28 03:23:11 PM UTC 2022 Downloading and extracting kube-burner binary
+ curl -LsS https://github.com/cloud-bulldozer/kube-burner/releases/download/v0.15.4/kube-burner-0.15.4-Linux-x86_64.tar.gz
+ tar xz
+ ./kube-burner init -c http-perf.yml --uuid=4576b26d-d972-4cea-bebe-875880cabf5f
INFO[2022-04-28 11:23:22] 🔥 Starting kube-burner (0.15.4@41a92affdd0b2c6e1299b94679e7f5c1e7e40cac) with UUID 4576b26d-d972-4cea-bebe-875880cabf5f
FATA[2022-04-28 11:23:22] Error rendering configuration template: rendering error: template: :3:26: executing "" at <.TERMINATIONS>: map has no entry for key "TERMINATIONS"
```

PR #389, opened 2022-04-28
Revert "Template out infrastructure to match with TERMINATIONS"

dry923:
Reverts cloud-bulldozer/e2e-benchmarking#363

This resulted in this issue: https://github.com/cloud-bulldozer/e2e-benchmarking/issues/388 . Reverting until we can debug


sjug (approved):
Thanks @dry923

PR #390, opened 2022-04-29
Template out infrastructure to match with TERMINATIONS

rsevilla87:
### Description

Copy of https://github.com/cloud-bulldozer/e2e-benchmarking/pull/363 but now exporting the TERMINATIONS env var


sjug (approved):
/lgtm & works

PR #391, opened 2022-04-29
Replace naked pods by deployments

rsevilla87:

### Description

Also leveraging kube-burner's mechanism to map environment variables to improve the code.


workloads/router-perf-v2/templates/nginx-pod.yml
```diff
@@ -1,18 +1,25 @@
-apiVersion: v1
-kind: Pod
+apiVersion: apps/v1
```
sjug, line 1: `templates/nginx-deploy.yml` already exists which is a deployment of nginx pods.

workloads/router-perf-v2/http-perf.yml
```diff
@@ -10,17 +10,17 @@ jobs:
   waitWhenFinished: true
   objects:
     - objectTemplate: templates/nginx-pod.yml
-      replicas: ${NUMBER_OF_ROUTES}
+      replicas: {{ .NUMBER_OF_ROUTES }}
```
sjug, line 13: `http-perf.yml` is now duplicating `http-perf-depl.yml`
The difference before was that `http-perf-depl.yml` was using deployments, and `http-perf.yml` was deploying pods. But you just changed the contents of the `templates/nginx-pod.yml` to be a deployment despite the filename and `templates/nginx-deploy.yml` already existing?

sjug (commented):
Pretty confused with your changes @rsevilla87

rsevilla87:
> Looks good @rsevilla87! One thing I thought could be helpful is to log the number of replicas & routes? (Especially since you disabled `set -x`)

I have added a small banner with all this information. BTW, I removed -x to prevent printing too much info in the log, (It makes a bit difficult to find something when something goes wrong in the benchmark)

Hope this banner provides enough info enough

sjug (approved):
Looks good, thanks @rsevilla87

PR #392, opened 2022-04-29
Reduce timeoutSeconds for startupProbe in node-density-cni

smalleni:
Does the same for node-density-cni-networkpolicy to improve granularity
of Pod Startup latency.


rsevilla87 (approved):
lgtm!

PR #393, opened 2022-04-30
adds benchmark-comparison to kube-burner

mohit-sheth:
This commits adds the ability to output a google spreadsheet URL with results

- Some example [touchstone configuration files](https://github.com/mohit-sheth/plow/tree/ovn-sheet/workloads/kube-burner/touchstone_configs) are added which can be used to populate the sheets. 
- By default we display top 10 results for a particular query but it can be changed via `NUM_LINES`
- It can also query a particular namespace by having `TOUCHSTONE_NAMESPACE=your_ns` or as `openshift*` to query all the namespaces. By default it will query the SDN (openshiftsdn or ovnk) namespace
- the [csv_modifier.py](https://github.com/mohit-sheth/plow/blob/ovn-sheet/workloads/kube-burner/csv_modifier.py) takes in the csv generated by touchstone and:  
  1. Sorts the data  
  2. Makes the data human readable  (can be improved)
  3. Outputs only top n rows  
  4. Appends all the different csv into one 

I have also added another metrics profile which is the same as metrics.yaml but has additional podCPU and podMemory metrics. 

TODO
remove timespamp from clusterversion query
pull gsheet_key from the vault


workloads/kube-burner/env.sh
```diff
@@ -7,7 +7,6 @@ export POD_READY_TIMEOUT=${POD_READY_TIMEOUT:-180}
 
 # Indexing variables
 export INDEXING=${INDEXING:-true}
-export ES_SERVER=${ES_SERVER:-https://search-perfscale-dev-chmf5l4sh66lvxbnadi4bznl3a.us-west-2.es.amazonaws.com:443}
```
paigerube14, line 10: did you mean to remove this line?

workloads/kube-burner/env.sh
```diff
@@ -69,3 +70,7 @@ export POD_READY_THRESHOLD=${POD_READY_THRESHOLD:-5000ms}
 # Alerting
 export PLATFORM_ALERTS=${PLATFORM_ALERTS:-false}
 
+# Output and Comparisons
+export COMPARISON_CONFIG=${COMPARISON_CONFIG:-""}
```
paigerube14, line 74: may be nice to give an example of how you can give multiple configs here. I would think that almost always it would be good to have use the files you added under the touchstone_configs folder. 

https://github.com/mohit-sheth/airflow-kubernetes/blob/1cba9bcbaab5f5ed2425e5b5a2bf85e45667a4ae/dags/openshift_nightlies/config/benchmarks/medium-control-plane.json#L49

workloads/kube-burner/metrics-profiles/metrics-ovn.yaml
```diff
@@ -0,0 +1,137 @@
...
+- query: sum(irate(apiserver_request_total{apiserver="kube-apiserver",verb!="WATCH"}[2m])) by (verb,resource,code) > 0
+  metricName: APIRequestRate
+
+# Kubeproxy and OVN service sync latency
+
+- query: histogram_quantile(0.99, sum(rate(kubeproxy_network_programming_duration_seconds_bucket[2m])) by (le)) > 0
+  metricName: serviceSyncLatency
+
+# Containers & pod metrics
+- query: sum(irate(container_cpu_usage_seconds_total{name!="",namespace=~"openshift-(etcd|oauth-apiserver|.*apiserver|ovn-kubernetes|sdn|ingress|authentication|.*controller-manager|.*scheduler|monitoring|logging|image-registry)"}[2m]) * 100) by (pod, namespace, node)
```
paigerube14, line 21: Looks like the only difference between this file and metrics.yaml is queries on line 21 and 24 are taken out. Should we just take these lines out in metrics.yaml?

utils/compare.sh
```diff
@@ -7,7 +7,7 @@ install_touchstone() {
   touchstone_tmp=$(mktemp -d)
   python3 -m venv ${touchstone_tmp}
   source ${touchstone_tmp}/bin/activate
-  pip3 install -qq git+https://github.com/cloud-bulldozer/benchmark-comparison.git
+  pip3 install -qq git+https://github.com/mohit-sheth/benchmark-comparison.git
```
paigerube14, line 10: Should we get any changes that are on your fork into the main repo of benchmark-comparison, so we don't need this line change and keep up to date with other changes that get into the main repo?

workloads/kube-burner/csv_modifier.py
```diff
@@ -0,0 +1,47 @@
+import pandas as pd
+from typing import *
+import sys
+import argparse
+import os
+
+def get_args() -> argparse.Namespace:
```
paigerube14, line 7: Should we add this to the main utils folder to be able to be used by all the workloads

workloads/kube-burner/common.sh
```diff
@@ -166,17 +166,66 @@ snappy_backup() {
...
+  if [[ $? == 0 ]]; then 
+   export div_by=1048576
+  fi
+  echo $config | grep -i latency
+  if [[ $? == 0 ]]; then
+    export div_by=1000
+  fi
+}
+
+run_benchmark_comparison() {
```
paigerube14, line 190: Thoughts on moving this to a more common place for all workload types?

utils/compare.sh
mohit-sheth: https://github.com/cloud-bulldozer/benchmark-comparison/pull/62 was merged
changing it back to upstream master

workloads/kube-burner/metrics-profiles/metrics-ovn.yaml
mohit-sheth: yes, the idea is to modify it further in upcoming PRs
Also it limits us indexing only useful metrics. For most of the runs where we will not need podCPU/podMem

workloads/kube-burner/env.sh
mohit-sheth: I did not, thank you

workloads/kube-burner/common.sh
mohit-sheth: Good point, currently there are 5 workloads which use this function in slightly different way than other.
I will create an issue and we can track it in another PR as it requires modification in other workloads

workloads/kube-burner/env.sh
paigerube14: should we also add a value or two as a default?

mohit-sheth: My intention was to not do any post benchmark analysis by default.
Hence https://github.com/mohit-sheth/plow/blob/53a419383c35109d750580c8fe628703208a76d6/workloads/kube-burner/common.sh#L191

workloads/kube-burner/common.sh
```diff
@@ -166,17 +166,66 @@ snappy_backup() {
...
+   if [[ -n ${ES_SERVER} ]] && [[ -n ${COMPARISON_CONFIG} ]]; then
+     log "Installing touchstone"
+     install_touchstone
+     NUM_LINES=${NUM_LINES:-10}
+     get_network_type
+     export TOUCHSTONE_NAMESPACE=${TOUCHSTONE_NAMESPACE:-"$network_ns"}
+     res_output_dir="/tmp/${WORKLOAD}-${UUID}"
```
rsevilla87, line 197: Are we sure we need this TOUCHSTONE_NAMESPACE variable?, keep in mind that the generated spreadsheet has column filters so finding/selecting a specific namespace shouldn't be a big deal.

workloads/kube-burner/common.sh
```diff
@@ -166,17 +166,66 @@ snappy_backup() {
...
+  if [[ $? == 0 ]]; then
+    export div_by=1000
+  fi
+}
+
+run_benchmark_comparison() {
+   if [[ -n ${ES_SERVER} ]] && [[ -n ${COMPARISON_CONFIG} ]]; then
+     log "Installing touchstone"
+     install_touchstone
+     NUM_LINES=${NUM_LINES:-10}
```
rsevilla87, line 194: Why do you want to only display a specific number of lines?, why not the entire comparison?

paigerube14: It might be nice to put each comparison config data in a different tab/worksheet under the same sheet if we are going to be printing out the entire comparison

PR #396, opened 2022-05-06
Disable self-signed cert warnings

rsevilla87:

### Description

Disable ugly self-signed cert urllib3 warnings


sjug (approved):
Thanks @rsevilla87 that'll do it, I don't think we care about certificate verification in our usage.

PR #399, opened 2022-05-11
Re-schedule monitoring-stack to worker nodes

rsevilla87:

### Description

Re-scheduling the monitoring stack to worker nodes before the benchmark starts will prevent the benchmark to be affected by prometheus noise (compaction, scraping, etc) and will help us to obtain more deterministic results.
The below graphs show an snippet of the system prometheus system activity during a benchmark

image

Where the highest CPU spikes happen during the database compaction: ~19:00 and 21:00

There's also a almost constant network I/O activity caused by metric scraping.

Another interesting fact is that we deploy 3 infrastructure nodes and there's only 2 prometheus replicas, so in some tests it's possible to have an ingress-controller in a node w/o any prometheus replica, that means that the test scenario can be different between runs.

### DIFF
--- a/workloads/router-perf-v2/common.sh
+++ b/workloads/router-perf-v2/common.sh
@@ -107,7 +107,7 @@ tune_liveness_probe(){
 reschedule_monitoring_stack(){
    [[ $(oc get cm -n openshift-monitoring cluster-monitoring-config --ignore-not-found --no-headers | wc -l) == 0 ]] && return 0
    log "Re-scheduling monitoring stack to ${1} nodes"
-   oc get cm -n openshift-monitoring cluster-monitoring-config -o yaml | sed "s#kubernetes.io/.*#kubernetes.io/${1}#g" | kubectl apply -f -
+   oc get cm -n openshift-monitoring cluster-monitoring-config -o yaml | sed "s#kubernetes.io/\w*#kubernetes.io/${1}#g" | oc apply -f -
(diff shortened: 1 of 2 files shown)


workloads/router-perf-v2/common.sh
```diff
@@ -104,6 +104,15 @@ tune_liveness_probe(){
   # TODO: validate liveness period and image
 }
 
+reschedule_monitoring_stack(){
+   log "Re-scheduling monitoring stack to ${1} nodes"
+   kubectl get cm -n openshift-monitoring  cluster-monitoring-config -o yaml | sed "s#kubernetes.io/.*#kubernetes.io/${1}#g" | kubectl apply -f -
```
sjug, line 109: `kubectl` commands?

rsevilla87: kubectl is also available in our airflow containers, fixed anyway

sjug:
@rsevilla87 Just thinking, why do we shuffle monitoring around? Why don't we just deploy it on worker nodes to begin with?

rsevilla87:
> @rsevilla87 Just thinking, why do we shuffle monitoring around? Why don't we just deploy it on worker nodes to begin with?

Other workloads require monitoring stack to be placed in infra nodes to avoid messing up with the workload itself. i.e networking (uperf) uses worker nodes to run the tests.

sjug (approved):
/lgtm

PR #400, opened 2022-05-11
Install specified version of ripsaw-cli

paigerube14:
### Description
It is confusing being able to set environment variables OPERATOR_REPO and OPERATOR_BRANCH when they aren't used to install the ripsaw-cli package at all, only the ripsaw operator apply. 

This will install the specified repo and branch depending on what env vars are set

### DIFF
--- a/utils/benchmark-operator.sh
+++ b/utils/benchmark-operator.sh
@@ -13,7 +13,7 @@ install_cli() {
       fi
   fi
   source ${ripsaw_tmp}/bin/activate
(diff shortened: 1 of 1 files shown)


paigerube14:
@jtaleric yes, will help with proper installation of benchmark-operator if a fork and branch is specified with environment variables

jtaleric:
ack @paigerube14 i just saw it wasn't updated but i see you fixed that. 😀

rsevilla87 (approved):
lgtm

PR #397, opened 2022-05-10
Max touchstone failures

rsevilla87:
Signed-off-by: Raul Sevilla <rsevilla@redhat.com>

### Description

With this change, touchstone will report error when there's more than 33% of failures on a comparison block, that is to say,  in the following comparison:

```
+-------------+--------+----------------------+-----------+---------------------+--------+-----------+------------+------------+
|  test_type  | routes | conn_per_targetroute | keepalive |       metric        | result | deviation |    4.9     | 4.11  |
+-------------+--------+----------------------+-----------+---------------------+--------+-----------+------------+------------+
| passthrough |  500   |          1           |     0     | avg(latency_95pctl) |  Pass  |  -2.55%   |   5571.5   |   5429.5   |
| passthrough |  500   |          1           |     1     | avg(latency_95pctl) |  Pass  |  -1.17%   |  148717.5  |  146976.5  |
| passthrough |  500   |          1           |    50     | avg(latency_95pctl) |  Pass  |  -0.36%   |   7680.0   |   7652.5   |
| passthrough |  500   |          20          |     0     | avg(latency_95pctl) |  Pass  |  -12.08%  |  81413.5   |  71577.0   |
| passthrough |  500   |          20          |     1     | avg(latency_95pctl) |  Pass  |  -1.84%   | 2903374.5  | 2849896.0  |
| passthrough |  500   |          20          |    50     | avg(latency_95pctl) |  Pass  |  -18.66%  |  132022.0  |  107385.5  |
| passthrough |  500   |          80          |     0     | avg(latency_95pctl) |  Fail  |  85.58%   |  446719.5  |  829003.0  |
| passthrough |  500   |          80          |     1     | avg(latency_95pctl) |  Pass  |  -19.71%  | 14254016.0 | 11444703.5 |
| passthrough |  500   |          80          |    50     | avg(latency_95pctl) |  Pass  |  -38.49%  | 1203344.0  |  740156.0  |
+-------------+--------+----------------------+-----------+---------------------+--------+-----------+------------+------------+
+-----------+--------+----------------------+-----------+--------------------------+--------+-----------+----------+-----------+
| test_type | routes | conn_per_targetroute | keepalive |          metric          | result | deviation |   4.9    | 4.11 |
+-----------+--------+----------------------+-----------+--------------------------+--------+-----------+----------+-----------+
| reencrypt |  500   |          1           |     0     | avg(requests_per_second) |  Pass  |   3.68%   | 122225.0 | 126720.5  |
[...]

PR #405, opened 2022-05-23
Add latency p95 back to router results comparison

rsevilla87:
@sjug


sjug (approved):
Thanks, restores touchstone config from https://github.com/cloud-bulldozer/e2e-benchmarking/pull/397

PR #407, opened 2022-05-24
Index cluster name

rsevilla87:

### Description

For some reason we were collecting the cluster_name and no indexing it.


mkarg75:
that's a clear case of "oops" - LGTM!

PR #408, opened 2022-05-24
Comparison format

rsevilla87:

### Description

In some cases we don't need to generate a CSV/google spreadsheet and just output the fancy benchmark-comparison output


sjug (approved):
/lgtm

PR #409, opened 2022-05-25
Add ROSA capabilities to upgrade_perf scripts

morenod:
Running upgrade_perf against a ROSA cluster will execute specific script

### DIFF
--- a/workloads/upgrade-perf/rosa_upgrade.sh
+++ b/workloads/upgrade-perf/rosa_upgrade.sh
@@ -1,0 +1,141 @@
+#!/usr/bin/env bash
+export rosa_environment=${ROSA_ENVIRONMENT:=staging}
+# Nightly versions cannot be upgraded, only valid channels for upgrading are stable and candidate
+export rosa_version_channel=${ROSA_VERSION_CHANNEL:=stable}
+_es_index=${ES_INDEX:=managedservices-timings}
+export control_plane_waiting_iterations=${ROSA_CONTROL_PLANE_WAITING:=90}
+export waiting_per_worker=${ROSA_WORKER_UPGRADE_TIME:=5}
+
+rosa_upgrade(){
+  if [ ${rosa_version_channel} == "nightly" ] ; then
+    echo "ERROR: Invalid channel group. Nightly versions cannot be upgraded. Exiting..."
+    exit 1
+  fi
+  echo "ROSA Cluster: ${ROSA_CLUSTER_NAME}"
+  echo "ROSA Environment: ${rosa_environment}"
+  echo "ROSA Channel Group: ${rosa_version_channel}"
+  export HOME=${HOME:=$PWD}
+  rosa login --env=${rosa_environment} --token=${ROSA_TOKEN}
+  if [ $(rosa list clusters | awk '{print $2}' | sed -e 1d | grep ${ROSA_CLUSTER_NAME}) != ${ROSA_CLUSTER_NAME} ] ; then
+    echo "ERROR: Cluster ${ROSA_CLUSTER_NAME} not found on rosa list clusters results. Exiting..."
+    exit 1
+  fi
+  if [ ! -z ${LATEST} ] && [ -z ${TOVERSION} ] ; then
+    echo "INFO: Getting latest version available to upgrade..."
+    VERSION=$(rosa list upgrade -c ${ROSA_CLUSTER_NAME} | sed -e 1d | head -1 | awk '{print $1}')
+  elif [ ! -z ${TOVERSION} ] && [ -z ${LATEST} ] ; then
+    echo "INFO: Checking if ${TOVERSION} version is available to upgrade..."
+    VERSION=$(rosa list upgrade -c ${ROSA_CLUSTER_NAME} | sed -e 1d | awk '{print $1}' | grep ${TOVERSION})
+  elif [ -z ${TOVERSION} ] && [ -z ${LATEST} ] ; then
+    echo "ERROR: Only LATEST or TOVERSION can be defined. Exiting..."
+    exit 1
+  else
+    echo "ERROR: Missing upgrade strategy. No LATEST or TOVERSION are defined. Exiting..."
+    exit 1
+  fi
+  if [ -z ${VERSION} ] ; then
+    echo "ERROR: No version to upgrade found running rosa list upgrade -c ${ROSA_CLUSTER_NAME}"
+    exit 1
+  else
+    echo "INFO: Upgrading cluster to ${VERSION} version..."
+  fi
+  CURRENT_VERSION=$(oc get clusterversion | grep ^version | awk '{print $2}')
+  rosa upgrade cluster -c ${ROSA_CLUSTER_NAME} --version=${VERSION} -y --schedule-date $(date +%Y-%m-%d) --schedule-time $(date -u --date="+6 minutes" +%H:%M)
+  # Sleep 6 minutes, rosa upgrade dont let to schedule an upgrade in less than 5 minutes from now
+  sleep 360
+  oc delete pods -n openshift-insights --all
+  oc delete pods -n openshift-managed-upgrade-operator --all
+  rosa_control_plane_upgrade_active_waiting ${VERSION}
+  if [ $? -eq 0 ] ; then
+    CONTROLPLANE_UPGRADE_RESULT="OK"
+  else
+    CONTROLPLANE_UPGRADE_RESULT="Failed"
+  fi
+  rosa_workers_active_waiting
+  if [ $? -eq 0 ] ; then
+    WORKERS_UPGRADE_RESULT="OK"
+  else
+    WORKERS_UPGRADE_RESULT="Failed"
+  fi
+  rosa_upgrade_index_results ${CONTROLPLANE_UPGRADE_DURATION} ${CONTROLPLANE_UPGRADE_RESULT} ${WORKERS_UPGRADE_DURATION} ${WORKERS_UPGRADE_RESULT} ${CURRENT_VERSION} ${VERSION}
+  exit 0
+}
+
+rosa_workers_active_waiting(){
+  start_time=$(date +%s)
+  WORKERS=$(oc get node --no-headers -l node-role.kubernetes.io/workload!="",node-role.kubernetes.io/infra!="",node-role.kubernetes.io/worker="" 2>/dev/null | wc -l)
+  # Giving waiting_per_worker minutes per worker
+  ITERATIONWORKERS=0
+  while [ ${ITERATIONWORKERS} -le $(( ${WORKERS}*${waiting_per_worker} )) ] ; do
+    if [ $(rosa describe cluster -c ${ROSA_CLUSTER_NAME} | grep ^Scheduled | wc -l) -ne 1 ] ; then
+      echo "INFO: Upgrade finished for ROSA, continuing..."
+      end_time=$(date +%s)
+      export WORKERS_UPGRADE_DURATION=$((${end_time} - ${start_time}))
+      return 0
+    else
+      LASTSTATUS=$(oc logs $(oc get pods -n openshift-managed-upgrade-operator -o Name | grep -v catalog) -n openshift-managed-upgrade-operator | grep "workers are upgraded" | tail -1)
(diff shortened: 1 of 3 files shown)


rsevilla87 (approved):
LGTM, (haven't reviewed ROSA specific code though)

workloads/upgrade-perf/rosa_upgrade.sh
```diff
@@ -0,0 +1,125 @@
...
+  ITERATIONS=0
+  while [ ${ITERATIONS} -le 90 ] ; do
+    VERSION_STATUS=($(oc get clusterversion | sed -e 1d | awk '{print $2" "$3" "$4}'))
+    if [ ${VERSION_STATUS[0]} == $1 ] && [ ${VERSION_STATUS[1]} == "True" ] && [ ${VERSION_STATUS[2]} == "False" ] ; then
+      # Version is upgraded, available=true, progressing=false -> Upgrade finished
+      echo "INFO: OCP upgrade to $1 is finished for OCP, now waiting for ROSA..."
+      rosa_workers_active_waiting
+      WORKERSUPGRADE=$?
+      end_time=$(date +%s)
+      export UPGRADE_DURATION=$((${end_time} - ${start_time}))
```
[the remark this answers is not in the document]
morenod: ok, now collecting both timers, control plane and workers

workloads/upgrade-perf/rosa_upgrade.sh
```diff
@@ -0,0 +1,125 @@
...
+  oc delete pods -n openshift-managed-upgrade-operator --all
+  # Giving 90 minutes for controlplane upgrade
+  start_time=$(date +%s)
+  ITERATIONS=0
+  while [ ${ITERATIONS} -le 90 ] ; do
+    VERSION_STATUS=($(oc get clusterversion | sed -e 1d | awk '{print $2" "$3" "$4}'))
+    if [ ${VERSION_STATUS[0]} == $1 ] && [ ${VERSION_STATUS[1]} == "True" ] && [ ${VERSION_STATUS[2]} == "False" ] ; then
+      # Version is upgraded, available=true, progressing=false -> Upgrade finished
+      echo "INFO: OCP upgrade to $1 is finished for OCP, now waiting for ROSA..."
+      rosa_workers_active_waiting
```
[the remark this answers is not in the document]
morenod: fixed

workloads/upgrade-perf/rosa_upgrade.sh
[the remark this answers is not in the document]
morenod: "controlplane_upgrade": "1515",
"workers_upgrade": "131",
"from_version": "4.10.10",
"to_version": "4.10.11",
"controlplane_upgrade_result": "OK",
"workers_upgrade_result": "OK",

morenod:
@dry923 @smalleni please check so I can continue working on integrating upgrades on airflow jobs

workloads/upgrade-perf/rosa_upgrade.sh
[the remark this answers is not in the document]
smalleni: Did it really only take 131 seconds to upgrade workers? That value seems a little suspect...

workloads/upgrade-perf/rosa_upgrade.sh
```diff
@@ -0,0 +1,138 @@
...
+  rosa upgrade cluster -c ${ROSA_CLUSTER_NAME} --version=${VERSION} -y --schedule-date $(date +%Y-%m-%d) --schedule-time $(date -u --date="+6 minutes" +%H:%M)
+  # Sleep 6 minutes, rosa upgrade dont let to schedule an upgrade in less than 5 minutes from now
+  sleep 360
+  oc delete pods -n openshift-insights --all
+  oc delete pods -n openshift-managed-upgrade-operator --all
+  rosa_control_plane_upgrade_active_waiting ${VERSION}
+  if [ $? -eq 0 ] ; then
+    CONTROLPLANE_UPGRADE_RESULT="OK"
+  else
+    CONTROLPLANE_UPGRADE_RESULT="Failed"
```
smalleni, line 45: I don't think the upgrade proceeds after the control plane upgrade fails, we should exit here since it doesnt make sense to wait for worker to update, since the worker MCPs only update after the control plane updates. We just need to construct the ES docuemnt here and exit 1 from the script.

workloads/upgrade-perf/rosa_upgrade.sh
```diff
@@ -0,0 +1,138 @@
...
+      LASTSTATUS=$(oc logs $(oc get pods -n openshift-managed-upgrade-operator -o Name | grep -v catalog) -n openshift-managed-upgrade-operator | grep "workers are upgraded" | tail -1)
+      echo "INFO: ${ITERATIONWORKERS}/$(( ${WORKERS}*5 )). Last Update: ${LASTSTATUS}."
+      echo "INFO: Waiting 60 seconds for the next check..."
+      ITERATIONWORKERS=$((${ITERATIONWORKERS}+1))
+      sleep 60
+    fi
+  done
+  echo "ERROR: ${ITERATIONWORKERS}/$(( ${WORKERS}*5 )). ROSA workers upgrade not finished after $(( ${WORKERS}*5 )) iterations. Exiting..."
+  end_time=$(date +%s)
+  export WORKERS_UPGRADE_DURATION=$((${end_time} - ${start_time}))
```
smalleni, line 78: So even if the worker upgrade fails - we are basically exporting the upgrade duration which is 5 mins * number of workers?

smalleni (commented):
LGTM, just couple of small comments that need to be addressed before merging.

workloads/upgrade-perf/rosa_upgrade.sh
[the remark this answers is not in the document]
morenod: Yes, it is ok, because the workers reboot starts on pararell with the masters and infra, and test is done o a 3 workers node

workloads/upgrade-perf/rosa_upgrade.sh
morenod: Actually, we should wait for the workers, because on that phase is where we check the result of the Rosa upgrade. On that phase, we know that we are waiting for the workers to be restarted, but what we are checking is the result of the Rosa show process, which is the scope of the script, to check the time that Rosa upgrade command takes to finish

workloads/upgrade-perf/rosa_upgrade.sh
morenod: Yep, that's right, but it always happen when you reach the timeout. You will send the timeout time, but the upgrade result will be fail, so we will exclude on the dashboards

workloads/upgrade-perf/rosa_upgrade.sh
smalleni: OK. As per my previous experience doing OCP upgrades, the workers are only upgraded after the control plane succeeds, maybe its different with ROSA?

workloads/upgrade-perf/rosa_upgrade.sh
```diff
@@ -0,0 +1,141 @@
...
+  else
+    echo "INFO: Upgrading cluster to ${VERSION} version..."
+  fi
+  CURRENT_VERSION=$(oc get clusterversion | grep ^version | awk '{print $2}')
+  rosa upgrade cluster -c ${ROSA_CLUSTER_NAME} --version=${VERSION} -y --schedule-date $(date +%Y-%m-%d) --schedule-time $(date -u --date="+6 minutes" +%H:%M)
+  # Sleep 6 minutes, rosa upgrade dont let to schedule an upgrade in less than 5 minutes from now
+  sleep 360
+  oc delete pods -n openshift-insights --all
+  oc delete pods -n openshift-managed-upgrade-operator --all
+  rosa_control_plane_upgrade_active_waiting ${VERSION}
```
dry923, line 46: I assume the upgrade waits till after the deleted pods are respawned and in good health. If that assumption is true we should wait till after they are up and running before calling rosa_control_plane_upgrade_active_waiting as that effectively starts the timer on the upgrade.

morenod: Upgrade starts automatically right after the pods are recreated, which happens just after deleting them.

Adding another active waiting just for them is adding complexity to the code with no benefit

workloads/upgrade-perf/rosa_upgrade.sh
```diff
@@ -0,0 +1,139 @@
...
+    WORKERS_UPGRADE_RESULT="Failed"
+  fi
+  rosa_upgrade_index_results ${CONTROLPLANE_UPGRADE_DURATION} ${CONTROLPLANE_UPGRADE_RESULT} ${WORKERS_UPGRADE_DURATION} ${WORKERS_UPGRADE_RESULT} ${CURRENT_VERSION} ${VERSION}
+  exit 0
+}
+
+rosa_workers_active_waiting(){
+  start_time=$(date +%s)
+  WORKERS=$(oc get node --no-headers -l node-role.kubernetes.io/workload!="",node-role.kubernetes.io/infra!="",node-role.kubernetes.io/worker="" 2>/dev/null | wc -l)
+  # Giving 5 minutes per worker
```
dry923, line 65: Can we variabilize the 5 min per worker?

workloads/upgrade-perf/rosa_upgrade.sh
```diff
@@ -0,0 +1,139 @@
...
+}
+
+rosa_upgrade_index_results(){
+  METADATA=$(grep -v "^#" << EOF
+{
+"uuid" : "${_uuid}",
+"platform": "ROSA",
+"cluster_name": "${ROSA_CLUSTER_NAME}",
+"network_type": "$(oc get network cluster -o json 2>/dev/null | jq -r .status.networkType)",
+"controlplane_upgrade": "$1",
```
dry923, line 121: Can this and the worker_upgrade one be appended with duration? ie workers_upgrade_duration?

workloads/upgrade-perf/rosa_upgrade.sh
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env bash
+export rosa_environment=${ROSA_ENVIRONMENT:=staging}
+# Nightly versions cannot be upgraded, only valid channels for upgrading are stable and candidate
+export rosa_version_channel=${ROSA_VERSION_CHANNEL:=stable}
+_es_index=${ES_INDEX:=managedservices-install-timings}
```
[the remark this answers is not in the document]
morenod: we have only one index for installing and upgrading, so we can see both results on the same board. I will rename it to something without including install or upgrade

workloads/upgrade-perf/rosa_upgrade.sh
morenod: ok done

workloads/upgrade-perf/rosa_upgrade.sh
```diff
@@ -0,0 +1,139 @@
...
+  echo "ERROR: ${ITERATIONWORKERS}/$(( ${WORKERS}*5 )). ROSA workers upgrade not finished after $(( ${WORKERS}*5 )) iterations. Exiting..."
+  end_time=$(date +%s)
+  export WORKERS_UPGRADE_DURATION=$((${end_time} - ${start_time}))
+  oc logs $(oc get pods -n openshift-managed-upgrade-operator -o Name | grep -v catalog) -n openshift-managed-upgrade-operator
+  rosa describe cluster -c ${ROSA_CLUSTER_NAME}
+  return 1
+}
+
+rosa_control_plane_upgrade_active_waiting(){
+  # Giving 90 minutes for controlplane upgrade
```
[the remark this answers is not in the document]
morenod: ok done

workloads/upgrade-perf/rosa_upgrade.sh
morenod: done

workloads/upgrade-perf/rosa_upgrade.sh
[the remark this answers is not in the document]
morenod: moved to managedservices-timings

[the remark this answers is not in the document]
dry923: sounds good

PR #412, opened 2022-05-27
Remove default value for METRICS_PROFILE

rsevilla87:

### Description

If se set a default value, the METRICS_PROFILE variable is not overriden in the workload, because of the expression `METRICS_PROFILE=${METRICS_PROFILE:-metrics-profiles/metrics-aggregated.yaml}`

### Fixes

The metrics-aggregated profile is not being set in cluster-density tests


Harshith-umesh (approved):
LGTM

PR #414, opened 2022-05-31
fix aws keys

venkataanil:
Currently we are overriding AWS_ACCESS_KEY_ID and
AWS_SECRET_ACCESS_KEY with OsdCcsAdmin user's keys.

Newly created OsdCcsAdmin user's aws token is needed for only
ocm-api-load tool.

So instead of overriding existing aws keys, we can create
OsdCcsAdmin user keys with different variable names.


chentex (approved):
:+1:

chentex (approved):

mukrishn (approved):

PR #417, opened 2022-06-01
Optimize pod count query

rsevilla87:

### Description

Pod counting was taking a while in big clusters. This new method optimizes that mechanism, since it doesn't requires to execute `oc describe` over each node.

### DIFF
--- a/workloads/kube-burner/common.sh
+++ b/workloads/kube-burner/common.sh
@@ -95,9 +95,17 @@ run_workload() {
 
 find_running_pods_num() {
   pod_count=0
-  for n in ${WORKER_NODE_NAMES}; do
-    pods=$(oc describe ${n} | awk '/Non-terminated/{print $3}' | sed "s/(//g")
-    pod_count=$((pods + pod_count))
+  # The next statement outputs something similar to:
+  # ip-10-0-177-166.us-west-2.compute.internal:20
+  # ip-10-0-250-197.us-west-2.compute.internal:17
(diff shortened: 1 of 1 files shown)


workloads/kube-burner/common.sh
```diff
@@ -95,9 +95,17 @@ run_workload() {
...
 find_running_pods_num() {
   pod_count=0
-  for n in ${WORKER_NODE_NAMES}; do
-    pods=$(oc describe ${n} | awk '/Non-terminated/{print $3}' | sed "s/(//g")
-    pod_count=$((pods + pod_count))
+  # The next statement outputs something similar to:
+  # ip-10-0-177-166.us-west-2.compute.internal:20
+  # ip-10-0-250-197.us-west-2.compute.internal:17
+  # ip-10-0-151-0.us-west-2.compute.internal:19
+  NODE_PODS=$(kubectl get pods --field-selector=status.phase=Running -o go-template --template='{{range .items}}{{.spec.nodeName}}{{"\n"}}{{end}}' -A | awk '{nodes[$1]++ }END{ for (n in nodes) print n":"nodes[n]}')
```
dry923, line 102: This wont count any pods that are in pending state about to be deployed. We could go w/ just checking for non-Completed pods via 
```
kubectl get pods --field-selector=status.phase!=Completed -o go-template --template='{{range .items}}{{.spec.nodeName}}{{"\n"}}{{end}}' -A | awk '{nodes[$1]++ }END{ for (n in nodes) print n":"nodes[n]}'
```

PR #421, opened 2022-06-03
Replace create token with sa new-token

rsevilla87:

### Description

The command create token is not available yet in the latest stable oc binary (used by the airflow container)

https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/scripts/run_benchmark.sh#L24

The oc subcommand sa new-token is compatible with previous oc versions and it's behavior is similar


dry923 (approved):
LGTM!

PR #422, opened 2022-06-06
Multiple python venvs to resolve conflicts on parallel workload runs

vishnuchalla:
### Description
Bug fix to create multiple python virtual environments to resolve conflicts on parallel workload runs.

#### Before bug fix
Before this bug fix when triggered multiple workloads in parallel on a host, both the workload used `/tmp/ripsaw` as the virtual python environments. So in the final step of a workload execution when the garbage collection happens if one of the workload finishes fist, it removes the directory `/tmp/ripsaw` which was also being used by the other workload ultimately resulting in the following cleanup error. 

Relevant Screenshots :

- <a href="https://drive.google.com/file/d/1eQ4UubemRL_eMnwAwBhqXg9sQQlgis2j/view?usp=sharing"></a>
- <a href="https://drive.google.com/file/d/10DWGGzo12maAafCpftVkC6HmK4QGyXSW/view?usp=sharing"></a>

### Fix
In order to separate each execution in a unique manner we are creating separate virtual environment directories using a unique uuid to avoid conflicts between multiple parallel runs.

#### After bug fix
Relevant Screenshots:

- <a href="https://drive.google.com/file/d/1nCvmilil2jOJl7a8ixsQXNe55pvyw6ff/view?usp=sharing"></a>
- <a href="https://drive.google.com/file/d/1rP5fzSTVvSuLx9tg_k2t6mlMh4dSJt_1/view?usp=sharing"></a>

After this fix both the parallel runs ran and cleaned up successfully.

### DIFF
--- a/utils/benchmark-operator.sh
+++ b/utils/benchmark-operator.sh
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 
 install_cli() {
-  ripsaw_tmp=/tmp/ripsaw-cli
+  export UUID=${UUID:-`uuidgen`}
+  run_dir=/tmp/${UUID}
+  ripsaw_tmp=${run_dir}/ripsaw-cli
   mkdir -p ${ripsaw_tmp}
   if [[ ! -f ${ripsaw_tmp}/bin/activate ]]; then
       if [[ "${isBareMetal}" == "true" ]]; then
@@ -16,7 +18,7 @@ install_cli() {
 
 remove_cli() {
   deactivate
-  rm -rf ${ripsaw_tmp}
+  rm -rf ${run_dir}
 }
 
 ############################################################################
@@ -40,7 +42,7 @@ deploy_benchmark_operator() {
 remove_benchmark_operator() {
   source ${ripsaw_tmp}/bin/activate
   ripsaw operator delete --repo=${1} --branch=${2}
-  rm -rf ${ripsaw_tmp}
+  rm -rf ${run_dir}
   remove_cli
 }
 
--- a/utils/benchmark-operator.sh
+++ b/utils/benchmark-operator.sh
@@ -4,6 +4,7 @@ install_cli() {
   export UUID=${UUID:-`uuidgen`}
   run_dir=/tmp/${UUID}
   ripsaw_tmp=${run_dir}/ripsaw-cli
+  log "creating python virtual environment at path: ${ripsaw_tmp}"
   mkdir -p ${ripsaw_tmp}
   if [[ ! -f ${ripsaw_tmp}/bin/activate ]]; then
       if [[ "${isBareMetal}" == "true" ]]; then
@@ -17,7 +18,9 @@ install_cli() {
 }
 
 remove_cli() {
+  log "deactivating python virtual environment at path: ${ripsaw_tmp}"
   deactivate
+  log "cleaning up unique run directory at path: ${run_dir}"
   rm -rf ${run_dir}
 }
 
@@ -42,7 +45,6 @@ deploy_benchmark_operator() {
 remove_benchmark_operator() {
   source ${ripsaw_tmp}/bin/activate
   ripsaw operator delete --repo=${1} --branch=${2}
-  rm -rf ${run_dir}
   remove_cli
 }
 


mukrishn (approved):
LGTM

dry923:
@vishnuchalla Please also take a look at ci/run_ci.sh as it blindly removes /tmp/ripsaw-cli. We might be able to remove that line.

rsevilla87:
This PR won't fix parallel runs, as the operator's namespace is always benchmark-operator.

vishnuchalla:
> This PR won't fix parallel runs, as the operator's namespace is always benchmark-operator.

But this can be taken as an action item to support parallel executions in the same cluster as a long-term perspective. But this will definitely take good amount of effort in making code changes at multiple places and thoroughly testing each and every workload to ensure nothing breaks.

mukrishn:
> This PR won't fix parallel runs, as the operator's namespace is always benchmark-operator.

@rsevilla87 yes, this is not to support parallel runs on same cluster but to use same server to run  on multiple cluster. Sometimes we run multiple workload across different cluster from same machine/vm, will hit this run time conflict. Recent slack [convo](https://coreos.slack.com/archives/C031XCAGH3R/p1650458204916499).

vishnuchalla:
> @vishnuchalla Please also take a look at ci/run_ci.sh as it blindly removes /tmp/ripsaw-cli. We might be able to remove that line.

Thanks for pointing out. It had no effect on our test from the start itself. But i remove this code as a best practice

vishnuchalla:
> /rerun all

Reran all the tests on my host. Here is the attached report 

Results for e2e-benchmarking CI Tests
Workload                | Test                           | Result | Runtime  |
------------------------|--------------------------------|--------|----------|
router-perf-v2 | ./ingress-performance.sh | PASS | 00:10:17
etcd-perf | ./run_etcd_tests_fromgit.sh  | PASS | 00:03:16
scale-perf | ./run_scale_fromgit.sh  | PASS | 00:14:13
network-perf | ./run.sh | PASS | 00:06:22
kube-burner | WORKLOAD=cluster-density ./run.sh | PASS | 00:05:25
kube-burner | WORKLOAD=node-density ./run.sh | PASS | 00:05:12
kube-burner | WORKLOAD=node-density-heavy ./run.sh | PASS | 00:06:06
kube-burner | WORKLOAD=pod-density-heavy ./run.sh | PASS | 00:05:48
kube-burner | WORKLOAD=max-services ./run.sh | PASS | 00:04:44
kube-burner | WORKLOAD=max-namespaces ./run.sh | PASS | 00:05:23
upgrade-perf | ./run_upgrade_fromgit.sh  | PASS | 00:01:33

dry923 (approved):
LGTM!

utils/benchmark-operator.sh
```diff
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
 
 install_cli() {
-  ripsaw_tmp=/tmp/ripsaw-cli
+  export UUID=${UUID:-`uuidgen`}
```
rsevilla87, line 4: Why are we exporting UUID again?

vishnuchalla: Yes I could have simply done that. But by doing this way: https://github.com/cloud-bulldozer/e2e-benchmarking/pull/422/files#diff-75780417466f89cde880aac56410f858a717b735c2dfea4b4fd587ba13bb7027L4 what I am trying to achieve here is to have a same UUID across the run so that if there was a case where we end up debugging the run, using the UUID generated above we should be able to simply identify the tmp directory created for that specific run.

<a href="https://drive.google.com/file/d/1Gmeghqn385nZ2JFpJ1ufO2kQHxVCfHxm/view?usp=sharing"></a>

If you look at the screenshot here. The UUID generated in the start of the run which is due to this code change is being across everywhere throughout the run including the place where Elastic Search Results are getting published. So I felt it would be an added advantage. 

Correct me if my intuition was wrong. I am even okay to make the code change as you mentioned above if it is going to keep things simple.

vishnuchalla: Removing the UUID in the ripsaw-cli setup step as to avoid duplicate code because every benchmark that is intended to use an UUID already generates one at the start of its run. Thanks for letting me know. Because of this discussion I have come to know that only few of our workloads actually need an UUID.

rsevilla87 (approved):
LGTM

PR #423, opened 2022-06-07
add ocm api load job timeout

venkataanil:
ocm api load testing takes 6 to 7 hours for completion. So set
api-load job timeout to 8 hours so that ocm-api-load job won't
terminate ocp-api-load pod within 8 hours.


morenod (approved):
LGTM

PR #425, opened 2022-06-08
Decrease verbosity of kube-burner tests

rsevilla87:

### Description

We're printing too much useless info in the screen. This will drop the verbosity of the benchmark. 
Note that the kube-burner log level is changed to error by default.


morenod (approved):
LGTM

PR #429, opened 2022-06-13
Update workloads with multiarch images for ARM testing support

svetsa-rh:
### Description
Updating images for workloads to support ARM in addition to x86. Changes for other workloads to follow.


rsevilla87 (approved):
LGTM!

PR #430, opened 2022-06-14
included few additional metrics for hypershift

mukrishn:


mukrishn:
@dry923  moved columns and color coded. Green till 80%, Yellow till 100, Orange till 120%, Red when 150% or above.

PR #434, opened 2022-06-24
Remove labeling nodes to execute workloads

morenod:
Labeling nodes is not authorized on rosa/ocm/osd clusters, so we need to remove it.

Removed also the hability of selecting workers nodes, as we are not using it, because all the workloads are being always executed on all the available workers.

Changed nodeSelector to affinity on all templates, because node selector do not allow negative queries, so there is no way to select workers but not workloads or infra


rsevilla87 (approved):
Big fan of this PR. LGTM!

PR #435, opened 2022-06-24
hypershift e2e script changes

mukrishn:
### Description

- Hypershift install args
- Hypershift CLI install inputs
- Indexing management cluster stats
- Firewall changes

### DIFF
--- a/workloads/hyper-scale/common.sh
+++ b/workloads/hyper-scale/common.sh
@@ -134,9 +134,12 @@ cleanup(){
 index_mgmt_cluster_stat(){
     echo "Indexing Management cluster stat..."
     echo "Installing kube-burner"
-    export KUBE_BURNER_RELEASE=${KUBE_BURNER_RELEASE:-0.16}
-    curl -L https://github.com/cloud-bulldozer/kube-burner/releases/download/v${KUBE_BURNER_RELEASE}/kube-burner-${KUBE_BURNER_RELEASE}-Linux-x86_64.tar.gz -o kube-burner.tar.gz
-    sudo tar -xvzf kube-burner.tar.gz -C /usr/local/bin/
+    KB_EXISTS=$(which kube-burner)
+    if [ $? -ne 0 ]; then
+        export KUBE_BURNER_RELEASE=${KUBE_BURNER_RELEASE:-0.16}
+        curl -L https://github.com/cloud-bulldozer/kube-burner/releases/download/v${KUBE_BURNER_RELEASE}/kube-burner-${KUBE_BURNER_RELEASE}-Linux-x86_64.tar.gz -o kube-burner.tar.gz
+        sudo tar -xvzf kube-burner.tar.gz -C /usr/local/bin/
+    fi
     export MGMT_CLUSTER_NAME=$(oc get infrastructure cluster -o jsonpath='{.status.infrastructureName}')
     export HOSTED_CLUSTER_NS="clusters-hypershift.*"
     envsubst < ../kube-burner/metrics-profiles/hypershift-metrics.yaml > hypershift-metrics.yaml
--- a/workloads/hyper-scale/common.sh
+++ b/workloads/hyper-scale/common.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -x
 
 source env.sh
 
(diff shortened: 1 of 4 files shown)


workloads/hyper-scale/env.sh
```diff
@@ -13,8 +13,20 @@ export NUMBER_OF_HOSTED_CLUSTER=${NUMBER_OF_HOSTED_CLUSTER:-2}
 export COMPUTE_WORKERS_NUMBER=${COMPUTE_WORKERS_NUMBER:-24}
 
 # Hosted cluster spec
+export HYPERSHIFT_OPERATOR_IMAGE="quay.io/hypershift/hypershift-operator:latest"
 export RELEASE_IMAGE=""
```
chaitanyaenr, line 17: User should be able the release image and others using environment variables like @mukrishn and I discussed.

chaitanyaenr:
@dry923 @morenod Need your review on this PR when you get time please? We are looking at leveraging this script for chaos testing HyperShift. Thanks.

workloads/hyper-scale/common.sh
```diff
@@ -4,22 +4,28 @@ source env.sh
 
 prep(){
     if [[ -z $(go version) ]]; then
-        curl -L https://go.dev/dl/go1.17.6.linux-amd64.tar.gz -o go1.17.6.linux-amd64.tar.gz
-        tar -C /usr/local -xzf go1.17.6.linux-amd64.tar.gz
+        curl -L https://go.dev/dl/go1.18.2.linux-amd64.tar.gz -o go1.18.2.linux-amd64.tar.gz
```
chaitanyaenr, line 7: We might want to add make, jq, kubectl/oc as dependencies to the readme. Thoughts?

morenod (approved):
LGTM

workloads/hyper-scale/common.sh
mukrishn: updated readme for make and jq, include cli installs in the script.

chaitanyaenr (approved):
Minor nit, other than that LGTM! Thanks @mukrishn!

workloads/hyper-scale/common.sh
```diff
@@ -1,25 +1,42 @@
...
+        tar -C /usr/local -xzf go1.18.2.linux-amd64.tar.gz
         export PATH=$PATH:/usr/local/go/bin
-        git clone --branch main https://github.com/openshift/hypershift
+    fi
+    if [[ ${HYPERSHIFT_CLI_INSTALL} != "false" ]]; then
+        echo "Remove current Hypershift CLI directory.."
+        sudo rm -rf hypershift || true
+        sudo rm /usr/local/bin/hypershift || true
```
dry923, line 15: Can we have it use a local version of hypershift as to a) not cause conflicts if I am running two tests from the same host and b) not explode my in use copy of it if I am testing something just for this e2e run?

mukrishn: Yes, That is possible by setting  `HYPERSHIFT_CLI_INSTALL` variable to `false`, so it will not re-install the CLI and use the local version.

dry923: Per my slack message. To clarify, I did not mean so it uses my currently installed hypershift cli but copies down to a local/dynamic folder. ie if i'm running it from /home/dry/foo it creates /home/dry/foo/hypershift-cli/...

workloads/hyper-scale/common.sh
krishvoor: Can we consider leveraging `-o jsonpath`, with no additional dependencies on jq, Thoughts?

workloads/hyper-scale/common.sh
krishvoor: Sorry to jump in between, but maybe like this, to avoid conflicts w/ local clones of hypershift.

```
export HYPERSHIFT_GIT_REPO=`mktemp -d`
cd $HYPERSHIFT_GIT_REPO
git clone -q --depth=1 --single-branch --branch ${HYPERSHIFT_CLI_VERSION} ${HYPERSHIFT_CLI_FORK} 
pushd hypershift
.....
```

mukrishn: yes, thats right. @dry923 lets do it in a separate PR? we can workaround that by setting the cli install to `false` for now.

PR #411, opened 2022-05-26
Combining metrics profiles files

paigerube14:
### Description
Want to combine the metrics and metrics-ovn yaml files so the default metrics file has podCPU and podMemory that can be used when creating the benchmark-comparison for kube-burner

### Fixes

1. Needing to set metrics-ovn.yaml file to be able to use all of the touchstone-config files for google sheet comparison generation
2. Remove duplication of files that are very similar

PR #438, opened 2022-06-29
Combine metric profiles

rsevilla87:
### Description

Combine metrics-aggregated.yaml and metrics.yaml files.

I managed to combine them thanks to expressions such as:
```
irate(process_cpu_seconds_total{service="kubelet",job="kubelet"}[2m]) * 100 and on (node) topk(3,avg_over_time(irate(process_cpu_seconds_total{service="kubelet",job="kubelet"}[2m])[{{ .elapsed }}:]) and on (node) kube_node_role{role="worker"})
```

Which uses the [magic variable {{ .elapsed }}](https://kube-burner.readthedocs.io/en/latest/metrics/#using-the-elapsed-variable) which takes the value of the benchmark duration. Now, in large-scale deployments we won't collect metrics for all worker nodes (which is overkill) but instead we'll collect metrics for the 3 worker nodes with more kubelet CPU usage, same for other similar metrics.


Harshith-umesh:
@rsevilla87 thoughts on reviewing and updating this PR's logic https://github.com/cloud-bulldozer/e2e-benchmarking/pull/411 here?

afcollins (approved):

Harshith-umesh (commented):
LGTM

rsevilla87:
Let's put this one on hold, I have to make some change in the grafana report to ensure we don't loss any interesting metric

PR #440, opened 2022-06-29
hypershift: updated kube-burner doc

### LINKED ISSUE #439: Missing documentation for cluster-density-ms workload
There is no documentation for the kube-burner cluster-density-ms workload. We should define what it creates and the variables that it can handle in the readme.

mukrishn:
### Description
Updated Readme of kube-burner workload for `cluster-density-ms`


dry923 (approved):
LGTM!

PR #443, opened 2022-07-11
enable system metrics collection for uperf

mohit-sheth:
Defaults to false


vishnuchalla:
LGTM! verified with: https://github.com/cloud-bulldozer/benchmark-operator/blob/master/e2e/uperf/uperf.yaml

workloads/network-perf/common.sh
```diff
@@ -11,6 +11,7 @@ export_defaults() {
   network_type=$(oc get network cluster -o jsonpath='{.status.networkType}' | tr '[:upper:]' '[:lower:]')
   export UUID=$(uuidgen)
   export CLUSTER_NAME=$(oc get infrastructure cluster -o jsonpath="{.status.infrastructureName}")
+  export PROM_TOKEN=$(oc sa get-token -n openshift-monitoring prometheus-k8s || oc sa new-token -n openshift-monitoring prometheus-k8s)
```
jtaleric, line 14: Ran some of this manually and saw : 
`Command "new-token" is deprecated, and will be removed in the future version. Use oc create token instead.`

rsevilla87:
@mohit-sheth do we actually need this?, we have thanos metrics...

mohit-sheth:
matching the timestamp can be a hassle, it is defaulted to false so maybe we only run it during readouts

workloads/network-perf/common.sh
mohit-sheth: looks like a 1.24/4.11 change, updated
Thanks

rsevilla87 (approved):
LGTM!

PR #444, opened 2022-07-12
Stop pinning uperf pods 

rsevilla87:

### Description

Once https://github.com/cloud-bulldozer/benchmark-operator/pull/781 lands, we won't need to pin uperf pods anymore.
Also removing the multi_az implementation, this functionality is useless when we have a fresh deployed cluster, since these cluster usually have 3 or 4 machinesets using different AZs each one.

Removing all this stuff helps in simplifying the benchmark implementation quite a bit.
In case we're interested in providing a pin node feature (which is not available in the orig implementation),  it be addressed in a different PR to keep the original purpose of this one


jtaleric (approved):
lgtm

afcollins (approved):
LGTM!

afcollins (approved):

PR #453, opened 2022-08-09
router-perf to exclude non-200 responses for latency 

mukrishn:
### Description
MB tool is failing to record the right `delay`(latency) for non-200 response and hence affecting the overall P99 and P90 latency value. The more non-200 response the worse P99 latency value(1657660884527 second) due to the issue, this fix will only include 200 response for latency calculation.

This is a temporary work around while we work on the new benchmark-operator.

### DIFF
--- a/workloads/router-perf-v2/workload.py
+++ b/workloads/router-perf-v2/workload.py
@@ -63,7 +63,9 @@ def run_mb(mb_config, runtime, output):
         if hit[2] not in result_codes:
             result_codes[hit[2]] = 0
         result_codes[hit[2]] += 1
-        latency_list.append(int(hit[1]))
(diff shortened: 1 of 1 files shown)


vishnuchalla (approved):
looks good to me.

smalleni (commented):
Were you also going to ship the no. of non-200 response codes to ES or do we already do it?

vishnuchalla:
> Were you also going to ship the no. of non-200 response codes to ES or do we already do it?

Looks like we already do it. Relevant code: https://github.com/cloud-bulldozer/e2e-benchmarking/blob/f3991ff2595f1571cc49f54598ee48c108da68b2/workloads/router-perf-v2/workload.py#L127

rsevilla87 (approved):
lgtm!

PR #455, opened 2022-08-18
Add new workload focussed on networking

smalleni:
This scenario emulates a popular customer usecase.

### DIFF
--- a/workloads/kube-burner/common.sh
+++ b/workloads/kube-burner/common.sh
@@ -22,6 +22,7 @@ export TOLERATIONS="[{key: role, value: workload, effect: NoSchedule}]"
 export UUID=${UUID:-$(uuidgen)}
 export OPENSHIFT_VERSION=$(oc version -o json | jq -r '.openshiftVersion') 
 export NETWORK_TYPE=$(oc get network.config/cluster -o jsonpath='{.status.networkType}') 
+export INGRESS_DOMAIN=$(oc get IngressController default -n openshift-ingress-operator -o jsonpath='{.status.domain}' || oc get routes -A --no-headers | head -n 1 | awk {'print$3'} | cut -d "." -f 2-)
 
 platform=$(oc get infrastructure cluster -o jsonpath='{.status.platformStatus.type}')
 
--- a/workloads/kube-burner/workloads/pods-service-route/pods-service-route.yml
+++ b/workloads/kube-burner/workloads/pods-service-route/pods-service-route.yml
@@ -1,0 +1,112 @@
+---
+global:
+  writeToFile: false
+  indexerConfig:
+    enabled: ${INDEXING}
+    esServers: ["${ES_SERVER}"]
+    insecureSkipVerify: true
+    defaultIndex: ${ES_INDEX}
+    type: elastic
+  measurements:
+    - name: podLatency
+      esIndex: ${ES_INDEX}
+{{ if eq .PPROF_COLLECTION "True" }}
+    - name: pprof
+      pprofInterval: {{ .PPROF_COLLECTION_INTERVAL }}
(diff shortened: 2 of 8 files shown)


workloads/kube-burner/common.sh
```diff
@@ -22,6 +22,7 @@ export TOLERATIONS="[{key: role, value: workload, effect: NoSchedule}]"
 export UUID=${UUID:-$(uuidgen)}
 export OPENSHIFT_VERSION=$(oc version -o json | jq -r '.openshiftVersion') 
 export NETWORK_TYPE=$(oc get network.config/cluster -o jsonpath='{.status.networkType}') 
+export INGRESS_DOMAIN=$( oc get IngressController default -n openshift-ingress-operator -o jsonpath='{.spec.domain}')
```
jtaleric, line 25: Looking at a couple of deployments I have (and others), this doesn't always have a value... 

```
jtaleric at rhino in ~/dev/scratch/kconfig $ oc get IngressController default -n openshift-ingress-operator -o jsonpath='{.spec.domain}'
 at rhino in ~/dev/scratch/kconfig $ oc get IngressController default -n openshift-ingress-operator -o jsonpath="{.spec}" | jq
{
  "clientTLS": {
    "clientCA": {
      "name": ""
    },
    "clientCertificatePolicy": ""
  },
  "httpCompression": {},
  "httpEmptyRequestsPolicy": "Respond",
  "httpErrorCodePages": {
    "name": ""
  },
  "replicas": 1,
  "tuningOptions": {},
  "unsupportedConfigOverrides": null
}

jtaleric at rhino in ~/dev/scratch/kconfig $ oc get clusterversion
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.10.24   True        False         19d     Cluster version is 4.10.24

```

Should we validate this value is set before the script proceeds?

smalleni: That's surprising, I'm not sure why that field wouldn't be populated (it is on my OSD cluster)...Perhaps I should trust good old bash?

oc get routes -A --no-headers | head -n 1 | awk {'print$3'} | cut -d "." -f 2-

smalleni: Fixed

mohit-sheth:
@smalleni can we have 30 webserver deployments/ns
that way with 1000 namespaces we can have a total of 40k pods/10k svc/10k routes?

smalleni:
> @smalleni can we have 30 webserver deployments/ns that way with 1000 namespaces we can have a total of 40k pods/10k svc/10k routes?

This is directly from the test that the customer is running though - do you ask because you want to do 40k pods with less than 200 nodes?

workloads/kube-burner/workloads/pods-service-route/pods-service-route.yml
```diff
@@ -0,0 +1,112 @@
...
+
+      - objectTemplate: webserver-service.yml
+        replicas: 10
+      
+      - objectTemplate: webserver-route.yml
+        replicas: 10
+        inputVars:
+          ingressDomain: ${INGRESS_DOMAIN}
+
+  - name: curl-pod-create
```
afcollins, line 88: Does this unique job run in sequence then, after the above completes?

afcollins (approved):
LGTM

dry923 (approved):
LGTM!

workloads/kube-burner/workloads/pods-service-route/pods-service-route.yml
smalleni: Yes, we need all the routes across all namespaces to be created because each pod in a namespace curls a random route across any of the namespaces

jtaleric (approved):
lgtm

PR #461, opened 2022-08-26
Adding churn ability for kube-burner workloads

dry923:
### Description

This aims to add the ability to "churn" deployments or namespaces after a workload has finished but prior to it being cleaned up. 
It takes in:
-  a duration of time to churn for
- a duration of time to wait between each round of churning
- a percentage of the JOB_ITERATIONS that it will churn
- a true|false flag to denote if we should churn
- a churn type of either pods or namespaces

After a workload has returned from the run_benchmark function we check if the churn flag is enabled. If so, we calculate the number of objects we need to churn based on a percentage of JOB_ITERATIONS. Every X seconds we then randomly pick deployments to scale down to 0 then back up to their previous value OR namespaces to delete/recreate (including objects in the namespace) until we hit our duration.


workloads/kube-burner/common.sh
```diff
@@ -242,3 +246,53 @@ prep_networkpolicy_workload() {
...
+  CHURN_DURATION="${CHURN_DURATION:-10}"
+  churn_end_time=$((${churn_start} + ${CHURN_DURATION}*60))
+  CHURN_WAIT="${CHURN_WAIT:-60}"
+  CHURN_PERCENT="${CHURN_PERCENT:-10}"
+
+  log "Churn duration: ${CHURN_DURATION} minutes"
+  log "Churn wait duration: ${CHURN_WAIT} seconds"
+  log "Churn percentage: ${CHURN_PERCENT}%"
+
+  namespace_arry=(`kubectl get namespaces -l kube-burner-uuid=${UUID} | grep -v NAME | awk '{print $1}'`)
```
jtaleric, line 263: `--no-headers`

workloads/kube-burner/common.sh
```diff
@@ -242,3 +246,70 @@ prep_networkpolicy_workload() {
...
+
+  churn_start=`date +%s`
+  churn_end_time=$((${churn_start} + ${CHURN_DURATION}*60))
+
+  log "Churn duration: ${CHURN_DURATION} minutes"
+  log "Churn wait duration: ${CHURN_WAIT} seconds"
+  log "Churn percentage: ${CHURN_PERCENT}%"
+  log "Churn type: ${CHURN_TYPE}"
+
+  namespace_arry=(`kubectl get namespaces -l kube-burner-uuid=${UUID} --no-headers | awk '{print $1}'`)
```
chentex, line 261: Should be `array`?

chentex (approved):
Just a small typo, but I don't know if it is because of a restricted keyword.

workloads/kube-burner/common.sh
dry923: It can be if you'd like. I just shortened it to save a key stroke :-)

morenod (approved):
LGTM

workloads/kube-burner/common.sh
dry923: @chentex updated to namespace_array

venkataanil (approved):
LGTM

mukrishn (approved):
LGTM

workloads/kube-burner/common.sh
```diff
@@ -242,3 +246,70 @@ prep_networkpolicy_workload() {
...
+	kubectl get -o yaml all -l kube-burner-uuid=${UUID} -n ${namespace_array[$rand_ns]} > churn_all.yaml
+
+	#Delete namespace
+        kubectl delete namespace ${namespace_array[$rand_ns]} > /dev/null
+
+	#Re-create objects
+	kubectl apply -f churn_namespace.yaml > /dev/null 2>&1
+	kubectl apply -f churn_configmaps.yaml > /dev/null 2>&1
+	kubectl apply -f churn_secrets.yaml > /dev/null 2>&1
+	kubectl apply -f churn_all.yaml > /dev/null 2>&1
```
smalleni, line 295: the `oc get all` command gets deployments and replicasets as well - a deployment naturally creates a replicaset - have noticed any issue with explicitly trying to create the replicast?

dry923: The replicasets that are created from the deployment does not have the kube-burner-uuid label on them. Thus, when I run the get all with -l kube-burner-uuid=${UUID} it leaves out the replica set

workloads/kube-burner/common.sh
```diff
@@ -242,3 +246,70 @@ prep_networkpolicy_workload() {
...
+        rand_rep=$(($RANDOM%${#rep_sets[@]}))
+        #set to 0 replicas; it will automatically get swapped back to the appropriate number of replicas
+        kubectl scale --replicas 0 replicaset ${rep_sets[$rand_rep]} -n ${namespace_array[$rand_ns]} > /dev/null &
+      elif [[ ${CHURN_TYPE} == "namespace" ]]; then
+	#Churn type is namespace
+	#Get all relevent configs
+	kubectl get -o yaml namespace ${namespace_array[$rand_ns]} > churn_namespace.yaml
+	kubectl get -o yaml configmaps -l kube-burner-uuid=${UUID} -n ${namespace_array[$rand_ns]} > churn_configmaps.yaml
+	kubectl get -o yaml secrets -l kube-burner-uuid=${UUID} -n ${namespace_array[$rand_ns]} > churn_secrets.yaml
+	kubectl get -o yaml all -l kube-burner-uuid=${UUID} -n ${namespace_array[$rand_ns]} > churn_all.yaml
```
rsevilla87, line 286: Why do we need get all?, you can do the same with just one command

`kubectl get -o yaml cm,secrets,deploy,route -l kube-burner-uuid=${UUID}`

workloads/kube-burner/README.md
```diff
@@ -189,6 +194,18 @@ Each iteration creates the following objects:
 - 20 secrets containing 2048 character random string
 - 10 configMaps containing a 2048 character random string
 
+### Churn
+
+Churning a workload allows you to scale down and then up a percentage of deployments or namespaces **AFTER** the workload has run, but before cleanup (if enabled). It takes in a percentage, **CHURN_PERCENT**, which is the percentage of JOB_ITERATIONS it churn during each churn cycle. After the appropriate number of deployments or namespaces have been churned it will sleep for the remainder of **CHURN_WAIT** seconds if the time it took to churn was less than the **CHURN_WAIT**. This loop continues until the **CHURN_DURATION** is reached.
+
+To churn 20% of your **JOB_ITERATIONS** using the namespace method every 30 seconds for a total duration (the calculated time from when the first churn begins + the **CHURN_DURATION**) of 60 minutes you would utilize these variables:
```
rsevilla87, line 201: I assume that this Churning feature doesn't work with node-density-* workloads, doesn't it?

workloads/kube-burner/common.sh
```diff
@@ -242,3 +246,70 @@ prep_networkpolicy_workload() {
...
+    for ((i=0; i<${modify_count}; i++)); do
+      #Pick random Namespace from the list
+      rand_ns=$(($RANDOM%${namespace_count}))
+
+      if [[ ${CHURN_TYPE} == "pod" ]]; then
+        rep_sets=(`kubectl get replicasets -n ${namespace_array[$rand_ns]} --no-headers | awk '{print $1}'`)
+        #Pick a random replicaset from the namespace
+        rand_rep=$(($RANDOM%${#rep_sets[@]}))
+        #set to 0 replicas; it will automatically get swapped back to the appropriate number of replicas
+        kubectl scale --replicas 0 replicaset ${rep_sets[$rand_rep]} -n ${namespace_array[$rand_ns]} > /dev/null &
```
rsevilla87, line 279: I think all the logic of the pod churn_type could be simplified with 
`oc delete pod --all -n ${namespace_array[$rand_ns]}`
 
These pods will be automatically re-created by their corresponding controllers

workloads/kube-burner/common.sh
smalleni: right, looks good then

workloads/kube-burner/README.md
dry923: It's not really designed that way today however I could add an option to do a % of the pods created in node-density. Do you want me to go that route or doc it as is today?

workloads/kube-burner/common.sh
dry923: I found when writing them all to the same yaml file it had issues re-deploying things as expected. Breaking into smaller groups worked every time.

workloads/kube-burner/common.sh
dry923: True, however deleting all the pods in the namespace would be likely more than what I have coded in at the moment. That said I am fine moving to this as an option.

workloads/kube-burner/README.md
rsevilla87: I think doc it should work fine for now

workloads/kube-burner/common.sh
rsevilla87: You mean that more lines would be required?, My suggestion is to replace all that block with just one line

dry923: Nope. I mean that the current implementation does a random replica set at a time and theres usually multiple of those per namespace in our cluster-density tests. You delete ALL the pods in a namespace while mines potentially fewer. Again, its splitting hairs and I'm fine deleting all the pods.

workloads/kube-burner/README.md
dry923: Added doc note

workloads/kube-burner/common.sh
dry923: Updated to use kubectl delete pod --all....

rsevilla87 (approved):
lgtm

PR #462, opened 2022-08-26
Enable node-density lite on windows workers

afcollins:
### Description
Scheduling labels, tolerations, and extra variables to customize node-density to run on specific nodes, in this case Windows workers.

See PERFSCALE-1848

### DIFF
--- a/workloads/kube-burner/run.sh
+++ b/workloads/kube-burner/run.sh
@@ -16,7 +16,7 @@ case ${WORKLOAD} in
   node-density)
     WORKLOAD_TEMPLATE=workloads/node-pod-density/node-pod-density.yml
     METRICS_PROFILE=${METRICS_PROFILE:-metrics-profiles/metrics.yaml}
-    NODE_COUNT=${NODE_COUNT:-$(kubectl get node -l node-role.kubernetes.io/worker,node-role.kubernetes.io/infra!=,node-role.kubernetes.io/workload!= -o name | wc -l)}
+    NODE_COUNT=${NODE_COUNT:-$(kubectl get node -l ${WORKER_NODE_LABEL},node-role.kubernetes.io/infra!=,node-role.kubernetes.io/workload!= -o name | wc -l)}
     PODS_PER_NODE=${PODS_PER_NODE:-245}
     label="node-density=enabled"
     label_node_with_label $label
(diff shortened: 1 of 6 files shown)


workloads/kube-burner/run.sh
```diff
@@ -15,8 +15,10 @@ case ${WORKLOAD} in
   ;;
   node-density)
     WORKLOAD_TEMPLATE=workloads/node-pod-density/node-pod-density.yml
+    INDEXING="false"
+    POD_NODE_SELECTOR=${POD_NODE_SELECTOR:-'{kubernetes.io/os: windows}'}
```
rsevilla87, line 19: You're setting these variables by default regardless you want to trigger the test against windows workers or not. That would affect the current behavior with Linux nodes.

rsevilla87 (changes requested):
Added some comments, I also miss documenting the new variables

workloads/kube-burner/run.sh
afcollins: Yes, you're right. That was a change for my local testing I can remove.

afcollins: removed it now

afcollins:
updated variable name and added documentation for new variables.

workloads/kube-burner/workloads/node-pod-density/node-pod-density.yml
```diff
@@ -81,5 +81,6 @@ jobs:
       - objectTemplate: pod.yml
         replicas: 1
         inputVars:
-          containerImage: gcr.io/google_containers/pause:3.1
+          containerImage: ${NODE_POD_DENSITY_IMAGE}
           nodeSelector: "${POD_NODE_SELECTOR}"
+          tolerations: "${POD_TOLERATIONS}"
```
rsevilla87, line 86: What is used this variable for? It doesn't have any default value

afcollins: It is the toleration for the `os=Windows:NoSchedule` taint added to the windows nodes and is default empty to it is not added to other workloads.
It is set from the [DAG](https://github.com/cloud-bulldozer/airflow-kubernetes/pull/250/files#diff-1ed8b3795e32ef8e399751c11b8b44dc549a696cf9640045dd025fd8e43bc7efR20).

Alternatively: Since it is only a toleration, we could leave it in place permanently without a variable.

rsevilla87: Yeah, I like your seccond approach, let's leave w/o a variable it as it doesn't affect the default behavior in linux nodes

afcollins: Sounds good. Removed it.
FWIW: the mcr image manifest appears to have linux and windows images, so we can do tests on both windows and linux workers in the same run.

rsevilla87 (approved):
lgtm!

PR #465, opened 2022-09-01
Remove 4 pairs scenario

rsevilla87:

### Description

Stop running this test with 4 pairs, it doesn't add any value to the results in our current scenarios, as we're already using the full bandwidth of the instance with 2 pairs. As shown in the picture below

image

The the graph above, 4 pairs have similar throughput compared to 2 pairs. Apart from this, using more than 1 pair is only required to observe how 2 different process (uperf) behave at the time of sharing the network link

i.e
2 pairs 1024B: 4870 x 2 = 9740
4 pairs 1024B: 2460 x 4 = 9840


dry923 (approved):
LGTM!

afcollins (approved):
LGTM

PR #466, opened 2022-09-01
Adjust fsync wal alerting

dry923:
### Description

We regularly see etcd disk wal fsync times between 10-20 ms without any visible detriment. To help avoid failures due to this aggressive time I'd like to move it up to 20ms.

### Fixes

Fixes mistype in description of backed commit duration


rsevilla87:
Why 20ms by default?, the case you reported was seen in ROSA based clusters, in that case we should create a specific alert-profile for ROSA clusters rather than modifying the default alert-profiles

dry923:
Looks like the latest 4.11 rosa run did not run into latency issues. I'll shelve this for now

PR #468, opened 2022-09-02
Improve alert profiles logic to be compatible with ROSA deployments

rsevilla87:
### Description

The logic should be able now to detect if we're benchmarking a ROSA cluster. I also have added a custom profile for ROSA with higher etcd latencies thresholds to prevent false positives going forward

### DIFF
--- a/workloads/kube-burner/alerts-profiles/cluster-density-rosa.yml
+++ b/workloads/kube-burner/alerts-profiles/cluster-density-rosa.yml
@@ -1,0 +1,1 @@
+cluster-density.yml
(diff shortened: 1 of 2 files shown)


dry923:
@rsevilla87 I think we can skip this for the time being. I confirmed that rosa moved to gp3 and per our last round of testing the results for etcd latencies were inline with what we would expect (<10ms)

rsevilla87:
> @rsevilla87 I think we can skip this for the time being. I confirmed that rosa moved to gp3 and per our last round of testing the results for etcd latencies were inline with what we would expect (<10ms)

Ok, I updated the PR to only add the ability to fetch the ROSA alert profile, since the current logic doesn't for ROSA based clusters. No metrics have been updated for now.

PR #473, opened 2022-09-07
Add case 1 network policy workload

jcaamano:
Scenario as per README using [NetworkPolicy test cases](https://docs.google.com/document/d/1OQtOXtqqHzV79vJ04r3z1icv5b-9y5AtWGrEgUMnvEg) as reference

Some caveats:
* Since there does not seem to be away to have a different number of replicas in each namespace, used an average of 20 replicas per namespace
* The case uses un-enforced egress rules to provide some information to the pod scraper, unfortunately in OVN there is a bug that causes the selected pods to be isolated for egress even if the policy type is not set to egress. Fixed being worked on here: https://github.com/ovn-org/ovn-kubernetes/pull/3162

Signed-off-by: Jaime Caamaño Ruiz <jcaamano@redhat.com>

### DIFF
--- a/workloads/kube-burner/workloads/networkpolicy/case1.yml
+++ b/workloads/kube-burner/workloads/networkpolicy/case1.yml
@@ -1,0 +1,96 @@
+---
+global:
+  writeToFile: false
+  indexerConfig:
+    enabled: ${INDEXING}
+    esServers: ["${ES_SERVER}"]
+    insecureSkipVerify: true
+    defaultIndex: ${ES_INDEX}
+    type: elastic
+  measurements:
+    - name: podLatency
+      esIndex: ${ES_INDEX}
+{{ if eq .PPROF_COLLECTION "True" }}
+    - name: pprof
+      pprofInterval: {{ .PPROF_COLLECTION_INTERVAL }}
+      pprofDirectory: /tmp/pprof-data
+      pprofTargets:
+
+      - name: kube-apiserver-cpu
+        namespace: "openshift-kube-apiserver"
+        labelSelector: {app: openshift-kube-apiserver}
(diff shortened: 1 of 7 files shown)


jcaamano:
@mohit PTAL

I did not change the overall strategy of the test but I have some comments about it:
* We are not waiting until the deny policy is enforced. Could a connection be potentially successful because it was attempted before  the deny policy is enforced?
* Should the connection to each pod be attempted in parallel on its own thread? Otherwise the time spent connecting with previous pods (including the time spent indexing their result) is never potentially accounted as time spent connecting to the former pod. If this is not needed because all the results are being aggregated and it is assumed that we are measuring the time it takes to connect to all pods in sequence, I still think that the time it takes to index the result of each connection is not being accounted for, so perhaps the timings need be aggregated and reported a single time at the end of the test.
* The result for a connection that was not ever successful is reported as taking a max timeout of 300s instead of reporting an error, is this intended?
* I have seen the test ending before each pod had a chance to attempt all the connections. How does kube-burner decide that a test is done?

workloads/kube-burner/workloads/networkpolicy/case1.yml
```diff
@@ -0,0 +1,96 @@
...
+
+      - name: etcd-cpu
+        namespace: "openshift-etcd"
+        labelSelector: {app: etcd}
+        cert: {{ .CERTIFICATE }}
+        key: {{ .PRIVATE_KEY }}
+        url: https://localhost:2379/debug/pprof/profile?timeout=30
+{{ end }}
+jobs:
+  - name: networkpolicy-case2
```
mohit-sheth, line 58: s/case2/case1

mohit-sheth (approved):
LGTM

PR #475, opened 2022-09-13
Remove benchmark-operator dependency

rsevilla87:

### Description

This is still a PoC, let's try to remove bechmark-operator dependency in kube-burner tests.
At the moment benchmark-operator is not actually adding any useful benefit to the control-plane tests, the main benefit it adds is that it takes care of deploying the kube-burner within the cluster, that guarantees API throughput and access, however, we normally execute our tests from a jump host that in almost 100% of cases has enough connectivity to access the cluster API at the given QPS/Burst params.
The metadata indexed by backpack doesn't add much value to the result, and we haven't never used it. With the metadata indexed by https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/utils/common.sh#L116-L123 should be enough.

Among the benefits of this this PR: we simplify the logic, hence less points of failure, no python needed in most cases, we improve debugging, reduce test duration...

**Some more changes are still required to make this PR 100% functional.**

I'd like to know your opinion first before moving forward @jtaleric @dry923 @paigerube14 @afcollins @mohit-sheth


paigerube14 (approved):
Looks good to me, used Qe's jenkins automation with this branch/fork and succeeds with no changes on our side.

rsevilla87:
> Looks good to me, used Qe's jenkins automation with this branch/fork and succeeds with no changes on our side.

great!, thanks for verifying!

rsevilla87:
@jtaleric @dry923 PTAL

dry923:
@rsevilla87 I'm fine with removing the operator as a dependency. If we want to keep the metadata collection it is still extremely simple to run it outside of the operator as an option as well. Thoughts?

jtaleric:
> @rsevilla87 I'm fine with removing the operator as a dependency. If we want to keep the metadata collection it is still extremely simple to run it outside of the operator as an option as well. Thoughts?

With kube-burner I am not sure we look at the metadata -- it does its own metadata collection in a way.  Maybe there are aspects it doesn't. 

overall lgtm.

dry923 (approved):
LGTM!

PR #477, opened 2022-09-14
Script to migrate cluster from OpenShiftSDN to OVNKubernetes

mukrishn:
### Description

- Steps are documented here - https://docs.google.com/document/d/1KrQsFpHIuFg5Lv03muY__0-MlwftMFqgEsiroCkrBA0/edit#heading=h.24easisenadj 
- Tested on ROSA cluster

PR #478, opened 2022-09-15
unlabel node only on cleanup

mukrishn:
### Description
Node unlabel only when cleanup is set, this let us use node-density to pre-load cluster before [migration](https://github.com/cloud-bulldozer/e2e-benchmarking/pull/477) or upgrade.


rsevilla87 (approved):
lgtm!

PR #479, opened 2022-09-21
adding set -e to kube-burner's run.sh

radez:
Errors are being obfuscated. Forcing them to bubble up.

### DIFF
--- a/workloads/kube-burner/run.sh
+++ b/workloads/kube-burner/run.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/bash
 
+set -e
+
 . common.sh
 . build_helper.sh
 


afcollins (approved):

rsevilla87:
> @rsevilla87 The catalyst for this change surfaced from looking at runs QE is doing outside of airflow. We don't want the tests to succeed if an error is thrown. Forcing an exit on error here will ensure that we don't get false positives on any run.
> 
> I imagine the output would be the same as it is now. We'll just get failures where errors have been being thrown that should have been aborting the tests.

oh, that's correct. Somehow I confused this parameter with `set -x`, to enable bash verbose mode... my bad.

The current implementation does `exit ${rc}  at the end of the execution, where rc is set depending on the benchmark CR result, find that logic at https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/utils/benchmark-operator.sh#L83-L84

So my question is, is there any corner case I'm missing you want to solve with this PR?

rsevilla87:
> the problem we're seeing is jq is not installed in the container QE is using in prow. The intention of raising errors right away was to catch this. maybe the error handling here isn't compatible with using set -e

indeed, set -e will break the current error handling implementation. I propose to restore bash options with `set +e` just after the "case" loop to avoid that

radez:
We've discussed going a different direction to fix our problem. I'll just close this out

PR #493, opened 2022-10-12
changes to hypershift metrics

mukrishn:
### Description

- Reduced the period for `min_over_time` queries
- updated node cpu usage to sum all cpu cores in the given range


dry923 (approved):
LGTM!

Issue #487, opened 2022-10-06
Router tests fail in Hypershift hosted cluster


chentex:
While running [router tests](https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/router-perf-v2/ingress-performance.sh) in an Hypershift hosted cluster execution of the tests fail running with the following error

```
Thu Oct  6 02:18:42 PM UTC 2022 Disabling cluster version and ingress operators
Error from server (NotFound): deployments.apps "cluster-version-operator" not found
```

After debugging with @dry923  the cluster we found out that effectively that deployment is not existing in the hosted cluster:

```
oc get all -n openshift-cluster-version
NAME                               TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
service/cluster-version-operator   ClusterIP   172.31.142.196   <none>        9099/TCP   38m
```

but it is running on the management cluster, in the namespace of the hosted cluster:

```
oc get pods -n clusters-vzepedam-hc
NAME                                              READY   STATUS    RESTARTS   AGE
....
cluster-version-operator-6f79ccc65f-gsm5j         1/1     Running   0          36m
.....
```

The issue is specifically happening when calling [this](https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/router-perf-v2/common.sh#L47)

Further investigation and analysis will be needed to solve this, because this issue will happen also when we try to access any of the components that are running in the Management cluster and not on the Hosted cluster.

PR #495, opened 2022-10-13
Modify router tests to work with HyperShift

smalleni:
Fixes: https://github.com/cloud-bulldozer/e2e-benchmarking/issues/487

### Description
This modifies the router tests to work with HyperShift Hosted Clusters since some of the control plane components (openshift-ingress-operator/openshift-cluster-version-operator) do not exist on the hsoted cluster and instead exist on a management cluster.

### DIFF
--- a/workloads/router-perf-v2/common.sh
+++ b/workloads/router-perf-v2/common.sh
@@ -4,6 +4,21 @@ source ../../utils/common.sh
 openshift_login
 source env.sh
 
+check_hypershift(){
+  if [[ "$(oc get infrastructure/cluster -o jsonpath='{.status.controlPlaneTopology}')" == "External" ]]; then
+    log "Detected that script is running on HyperShift Hosted Cluster"
+    HYPERSHIFT=true
+    export CLUSTER_NAME=$(oc get infrastructure/cluster -o json | jq -r '.status.etcdDiscoveryDomain' | awk -F. {'print$1'})
+    export HYPERSHIFT_HOSTED_KUBECONFIG="${KUBECONFIG}"
+    if [[ -z "${HYPERSHIFT_MANAGEMENT_KUBECONFIG}" ]]; then
+      log "Path of Management Cluster kubeconfig required in the case of HyperShift"
+      exit 1
+    fi
+  else
+    HYPERSHIFT=false
+  fi
+}
+    
 get_scenario(){
   # We consider a large scale scenario any cluster with more than the given threshold
   if [[ ${NUM_NODES} -ge ${LARGE_SCALE_THRESHOLD} ]]; then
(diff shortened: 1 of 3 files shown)


chentex (changes requested):
First part works correctly but by the end of the scrip it fails again, it is necessary to check again in this lines: https://github.com/cloud-bulldozer/e2e-benchmarking/pull/495/files#diff-88112b993948b739e6f3469b2cc02d704e1ad5c42932086bfe397b2c3b4a7a55R179-R183

smalleni:
> First part works correctly but by the end of the scrip it fails again, it is necessary to check again in this lines: https://github.com/cloud-bulldozer/e2e-benchmarking/pull/495/files#diff-88112b993948b739e6f3469b2cc02d704e1ad5c42932086bfe397b2c3b4a7a55R179-R183

Can you paste what part fails? Not sure I caught that.

workloads/router-perf-v2/common.sh
```diff
@@ -149,8 +178,12 @@ run_mb(){
...
 enable_ingress_operator(){
   log "Enabling cluster version and ingress operators"
-  oc scale --replicas=1 -n openshift-cluster-version deploy/cluster-version-operator
-  oc scale --replicas=1 -n openshift-ingress-operator deploy/ingress-operator
+  if "${HYPERSHIFT}"; then 
+    export KUBECONFIG="${HYPERSHIFT_MANAGEMENT_KUBECONFIG}"
+    oc scale --replicas=1 -n "${NAMESPACE}" deploy/cluster-version-operator
+    oc scale --replicas=1 -n "${NAMESPACE}" deploy/ingress-operator
+    export KUBECONFIG="${HYPERSHIFT_HOSTED_KUBECONFIG}"
+  fi
```
mukrishn, line 186: Now skipping regular cluster, missing an else condition i guess.

chentex, line 186: I think if we leave it like this it will only run this for `hypershift`. it is also needed to run for a normal cluster, in my tests I used this code to achieve it:

```suggestion
  if [[ ${HYPERSHIFT} ]]; then
    export KUBECONFIG=${HYPERSHIFT_MANAGEMENT_KUBECONFIG}
    NAMESPACE="clusters-${CLUSTER_NAME}"
  else
    NAMESPACE="openshift-cluster-version"
  fi
  log "Enabling cluster version and ingress operators"
  oc scale --replicas=${NUMBER_OF_ROUTERS} -n ${NAMESPACE} deploy/cluster-version-operator
  oc scale --replicas=${NUMBER_OF_ROUTERS} -n ${NAMESPACE} deploy/ingress-operator
  if "${HYPERSHIFT}"; then KUBECONFIG="${HYPERSHIFT_HOSTED_KUBECONFIG}"; fi
```

It is basically using your logic from the previous steps to run those two `scale` commands

chentex (changes requested):
Some changes to `enable_ingress_operator` so it runs correctly with all types of clusters.

mukrishn:
@smalleni Just nitbit and optional, an easiest way to switch `KUBECONFIG` by passing it in the CLI itself will over ride `env` value, so can avoid updating env variable again and again.  
ex:
```
oc scale --replicas=1 -n "${NAMESPACE}" deploy/cluster-version-operator --kubeconfig ${HYPERSHIFT_MANAGEMENT_KUBECONFIG}
```

workloads/router-perf-v2/common.sh
smalleni: fixed

chentex (approved):
LGTM

smalleni:
@mukrishn @dry923 can we merge this now?

mukrishn (approved):

PR #499, opened 2022-10-14
index platform for hypershift

mukrishn:
### Description
Hypershift hosted cluster doesn't have the tag and that set the platform to NULL.


dry923 (approved):
LGTM!!

PR #501, opened 2022-10-19
updated path for metric.yaml

mukrishn:
### Description
Updating the path of rendered `metrics.yml` to be dynamic using `KUBE_DIR` env var.


dry923 (approved):
LGTM!

PR #507, opened 2022-11-08
Conc builds update to use python3 

paigerube14:
### Description
I updated the build_test.py script in the svt repo to use python3 packages instead of python2. Need to update this script to take out python 2 references 

Also the eap imagestream name had been updated and the concurrent-build script was not able to properly run that type of build

### Fixes
`Error "Error resolving ImageStreamTag jboss-eap64-openshift:latest in namespace openshift: imagestreams.image.openshift.io "jboss-eap64-openshift" not found" for field "from".`

### DIFF
--- a/workloads/kube-burner/build_helper.sh
+++ b/workloads/kube-burner/build_helper.sh
@@ -25,15 +25,16 @@ function prepare_builds_file()
 function install_svt_repo() {
   rm -rf svt
   git clone --single-branch --branch ${build_test_branch} ${build_test_repo} --depth 1
+  pip3 install future pytimeparse
 }
 
 function run_builds() {
+  
   for i in "${build_array[@]}"
   do
     log "running $i $1 concurrent builds"
     fileName="conc_builds_$1.out"
-    pip install future pytimeparse
-    python svt/openshift_performance/ose3_perf/scripts/build_test.py -z -a -n 2 -r $i -f running-builds.json >> $fileName 2>&1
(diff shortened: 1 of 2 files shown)


workloads/kube-burner/build_helper.sh
```diff
@@ -25,20 +25,15 @@ function prepare_builds_file()
...
-  python2 -m pip install futures pytimeparse logging
 }
 
 function run_builds() {
   for i in "${build_array[@]}"
   do
     log "running $i $1 concurrent builds"
     fileName="conc_builds_$1.out"
-    python2 svt/openshift_performance/ose3_perf/scripts/build_test.py -z -a -n 2 -r $i -f running-builds.json >> $fileName 2>&1
+    pip install future pytimeparse
```
jtaleric, line 35: should we explicitly call `pip3` here?

workloads/kube-burner/build_helper.sh
```diff
@@ -25,20 +25,15 @@ function prepare_builds_file()
...
+    python svt/openshift_performance/ose3_perf/scripts/build_test.py -z -a -n 2 -r $i -f running-builds.json >> $fileName 2>&1
```
jtaleric, line 36: similar here, should we call `python3` here?

jtaleric (commented):
Some comments inline, just curious if we should explicitly call the python version we want/need.

jtaleric (approved):
lgtm

mohit-sheth (approved):
lgtm

PR #510, opened 2022-11-15
Remove reference to HYPERSHIFT_MANAGEMENT_KUBECONFIG on env.sh as emp…

morenod:
…ty, because if not, there is no way to set the var as env var

### Description

With current configuration, we can only set the var on the file, any setting from the env will be overrrided as ""


mukrishn (approved):

mukrishn:
Good to be like this?
```
export HYPERSHIFT_MANAGEMENT_KUBECONFIG=${HYPERSHIFT_MANAGEMENT_KUBECONFIG:-""}
```

morenod:
> Good to be like this?
> 
> ```
> export HYPERSHIFT_MANAGEMENT_KUBECONFIG=${HYPERSHIFT_MANAGEMENT_KUBECONFIG:-""}
> ```

Yeah, better option, updated

dry923 (approved):
LGTM!

PR #533, opened 2023-02-09
fleet manager prometheus metrics

venkataanil:
SRE is using these prometheus metrics in their grafana dashboards [1]. After fleet manager scale testing, we manually run kube-burner with this metric profile to index into our internal ES. We have internal grafana dashboard [2] which uses these metrics from our ES.

[1] https://grafana.app-sre.devshift.net/d/osd_fleet_manager1/osd-fleet-manager-metrics1 [2] https://grafana.rdu2.scalelab.redhat.com:3000/d/osd_fleet_manager1/osd-fleet-manager-metrics1

### DIFF
--- a/workloads/fleet-manager/metrics.yaml
+++ b/workloads/fleet-manager/metrics.yaml
@@ -1,0 +1,238 @@
+- query: 'sum(fleet_manager_management_cluster_hosted_cluster_count{ job="fleet-manager-metrics", namespace="osd-fleet-manager-stage",  region="us-east-2"}) by (management_cluster)'
+  metricName: managementClusterHostedClusterCount
+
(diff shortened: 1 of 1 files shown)


venkataanil:
Initially submitted this file in perfscale-managed-services repo with PR https://github.com/cloud-bulldozer/perfscale-managed-services/pull/110 . However I got suggestions to submit in e2e-benchmarking rather than perfscale-managed-services.

krishvoor (approved):
Lgtm

workloads/fleet-manager/metrics.yaml
```diff
@@ -0,0 +1,238 @@
+- query: 'sum(fleet_manager_management_cluster_hosted_cluster_count{ job="fleet-manager-metrics", namespace="osd-fleet-manager-stage",  region="us-east-2"}) by (management_cluster)'
```
jtaleric, line 1: Will it always be `us-east-2` ?

jtaleric: I would suggest just having this be a template where we use `envsub` to swap out the region

venkataanil: we got us-east-2 for our testing. I will update it to use envsub

PR #534, opened 2023-02-10
Remove workload labels from scale-openshift task

afcollins:
### Fixes
e2e-benchmarking#530
scale-ci-deploy#202
airflow-kubernetes#303


afcollins:
Closing, as there is a simpler env fix in e2e-benchmarking already.

https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/scale-perf/env.sh#L21

PR #536, opened 2023-02-17
Add nodeMemoryUtilization for infra and control plane nodes

afcollins:
### Description

Currently kube-burner runs only capture memoryUtilization for Workers, but it is helpful to have for all nodes.


rsevilla87:
Fixed at https://github.com/cloud-bulldozer/kube-burner/pull/359

PR #543, opened 2023-03-06
Fix total_pod_count check

rsevilla87:
### Description

Kube-burner does not run as a pod anymore, there's no need to substract 1 to the result.
Also, the check should be done after dividing total_pod_count by 2 when the workload is node-density-heavy/cni


jtaleric (approved):
/lgtm

PR #544, opened 2023-03-09
Fix total_pod_count check

rsevilla87:
### Description

The check
```
if [[ ${total_pod_count} -le 0 ]]; then
```

Must be run after calculating total_pod_count


rsevilla87:
Mistakenly dupe of #543

PR #546, opened 2023-03-09
Revert kube-burner binary to 0.17.3

afcollins:
### Description
Workload configs have not been updated to be compatible with v1.4.2. So reverting to 0.17 for now.

### Fixes
`[2023-03-09, 19:39:51 UTC] {subprocess.py:92} INFO - time="2023-03-09 19:39:51" level=fatal msg="Error decoding configuration file node-pod-density.yml: yaml: unmarshal errors:\n  line 3: field writeToFile not found in type config.GlobalConfig\n  line 12: field esIndex not found in type types.Measurement"`

### DIFF
--- a/workloads/kube-burner/common.sh
+++ b/workloads/kube-burner/common.sh
@@ -65,7 +65,11 @@ run_workload() {
     mv ${KUBE_BURNER_DIR}/bin/amd64/kube-burner ${KUBE_DIR}/kube-burner
     rm -rf ${KUBE_BURNER_DIR}
   else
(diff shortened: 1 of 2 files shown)


jtaleric (changes requested):
Maybe we default to `0.17.3` but if `METRICS_ENDPOINT` is set, we default to 1.4.2?

afcollins:
Updated, pls check. The changes still pull 0.17.3 for me (just verified) so I think they're good.

PR #549, opened 2023-03-22
Update logging workload to OLO 5.6

morenod:
Update code of logging workload to the docs published for 4.12 and OLO 5.6


workloads/logging/README.md
```diff
@@ -27,15 +33,19 @@ Default: 3
...
-Default: 'gp2'
+Default: 'gp3-csi'
 Storage class to use for the persistent storage. The faster the storage, better the Elasticsearch performance.
 
 ### ES_STORAGE_SIZE
 Default: `100G`
 Each data node in the cluster is bound to a Persistent Volume Claim that requests the size specified using this variable from the cloud storage.
 
+### ES_MEMORY_LIMITS
+Default: `16Gi`
```
dry923, line 44: It looks like the default request and limit are the same. Do we really even need to set a limit here?

workloads/logging/common.sh
```diff
@@ -14,9 +14,9 @@ log() {
 log "Checking if oc client is installed"
 
 # Check cluster's health
-if [[ ${CERBERUS_URL} ]]; then
-  response=$(curl ${CERBERUS_URL})
-  if [ "$response" != "True" ]; then
+if [[ "${CERBERUS_URL}" ]]; then
```
dry923, line 17: We can remove the entire Cerberus section. It is not used anymore

workloads/logging/common.sh
```diff
@@ -33,56 +33,56 @@ deploy_logging_stack() {
...
-  envsubst < files/log_generator.yaml > log_generator_$timestamp.yaml
-  if [[ ${DEPLOY_LOGGING} == "true" ]]; then
-    # Get bearer token and ES url if applicable
-    if [[ ${CUSTOM_ES_URL} == "" ]]; then
-      ES_BACKEND_TOKEN=`oc whoami -t`
-      ES_BACKEND_URL=`oc get route elasticsearch -n openshift-logging -o jsonpath={.spec.host}`
-    else
-      ES_BACKEND_URL=$CUSTOM_ES_URL
-    fi
+  envsubst < files/log_generator.yaml > temp/log_generator_"${timestamp}".yaml
```
dry923, line 41: does this create the temp dir if it doesnt exist?

workloads/logging/common.sh
morenod: removed

workloads/logging/common.sh
morenod: good catch, adding creation of the folder

workloads/logging/README.md
rsevilla87: We can keep this variable as long as defaults make sense, which is the case :)

workloads/logging/deploy_logging_stack.sh
```diff
@@ -44,7 +44,7 @@ function cleanup() {
   while [[ $( oc get projects | grep -w "openshift-logging\|openshift-operators-redhat") ]]; do
     sleep 5
     wait_time=$((wait_time+5))
-    if [[ $wait_time -ge $TIMEOUT ]]; then
+    if [[ "${wait_time}" -ge "${TIMEOUT}" ]]; then
```
rsevilla87, line 47: Can we can reduce this whole function into:
```shell
cleanup() {
  oc delete --wait=true project openshift-logging openshift-operators-redhat --ignore-not-found --timeout=${TIMEOUT}
}

File: workloads/logging/env.sh [hunk 5]
```diff
@@ -13,6 +13,7 @@ export DURATION=${DURATION:-1}
 export MESSAGES_PER_SECOND=${MESSAGES_PER_SECOND:-0}
 export POD_COUNT=${POD_COUNT:-1}
 export TIMEOUT=${TIMEOUT:-600}
+export DEBUG=${DEBUG:-true}
```
rsevilla87, line 16: We normally don't enable DEBUG mode by default

workloads/logging/deploy_logging_stack.sh
morenod: rewritted cleanup function to use the --wait

workloads/logging/deploy_logging_stack.sh
```diff
@@ -67,24 +67,15 @@ oc wait --for=condition=available -n openshift-logging deployment/cluster-loggin
...
-  wait_time=$((wait_time+5))
-  if [[ $wait_time -ge $TIMEOUT ]]; then
-    log "Fluentd daemonset is not ready after $TIMEOUT, please check. Exiting"
-    exit 1
-  fi
-done
 log "Logging stack is up"
 
-if [[ $CUSTOM_ES_URL == "" ]]; then
+if [[ "${CUSTOM_ES_URL}" == "" ]]; then
```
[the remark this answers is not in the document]
morenod: Changed

workloads/logging/env.sh
morenod: changed

workloads/logging/deploy_logging_stack.sh
```diff
@@ -37,22 +37,15 @@ function install() {
...
-  oc delete --wait=true project openshift-operators-redhat --ignore-not-found
-  while [[ $( oc get projects | grep -w "openshift-logging\|openshift-operators-redhat") ]]; do
-    sleep 5
-    wait_time=$((wait_time+5))
-    if [[ $wait_time -ge $TIMEOUT ]]; then
-      log "openshift-logging/openshift-operators-redhat namespaces still exists after $TIMEOUT, please check. Exiting"
-      exit 1
-    fi
-  done
+  if ! oc delete --wait=true project openshift-logging openshift-operators-redhat --ignore-not-found --timeout=180s; then
```
rsevilla87, line 41: Why don't you use the variable ${TIMEOUT} rather than hardcoded 180s?

workloads/logging/deploy_logging_stack.sh
```diff
@@ -37,22 +37,15 @@ function install() {
...
+    log "openshift-logging/openshift-operators-redhat namespaces exists after 3 minutes, please check. Exiting..."
```
rsevilla87, line 42: same as above

workloads/logging/deploy_logging_stack.sh
morenod: because we were using it wrong. TIMEOUT is the time of log_generator to wait for all messages to be stored, it is being used `[here](https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/logging/files/log_generator.yaml#L20)`. We set it to HOURS, we cannot use it for deleting objects.

Also, in this same file, we have another wait hardcoded to 180, `[here](https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/logging/deploy_logging_stack.sh#L68)`

rsevilla87: ok!, can you update the docs then?, According to the docs
```
### TIMEOUT
Default: 180

Time to wait for resources created/deleted before exiting
```

morenod: done

rsevilla87 (approved):
LGTM!

krishvoor (approved):
lgtm

workloads/logging/README.md
smalleni: Agree, request=limit makes this a guaranteed pod which has implications in the context of preemption - guaranteed pods are preempted last, so there's probably several reasons why we have requests=limit for ES.

dry923: I don't disagree that the values may be reasonable, etc. My thinking here was if we are testing the logging stack we shouldn't limit the backend data store as it is just something that we may hit (OOM of ES) which would negatively impact our tests but not be prudent to the results. I.e. if we cause ES to OOM it does NOT indicate a failed test. It just means we need to change those values and re-run. IMHO removing the limit just potentially saves us steps/debugging down the line.

PR #555, opened 2023-04-04
Index camelCase field for metadata

rsevilla87:
### Description
Add camelCase metadata fields to make it compatible with the OCP wrapper dashboards.


jtaleric (approved):
lgtm, i didn't cross reference the fields here with what kube-burner does, since there is some shared fields between the two here..

PR #556, opened 2023-04-11
Fix router large scale scenario

smalleni:
Currently we check for greater than or equal to 24 nodes excluding workload node to execute the large scenario. This means that even if you have a 24 worker cluster, you will actually end up running the small scale scenario since one node is excludded when counting total workers.

However, we still continue to use the workload node to place the backend nginx pods, which makes our logic of determining small/large scale senario and labeling the workload node counter-intuitive. The nginx backend pods are fighting for cpu time on the same node where mb client is firing requests off of.

This patch does the following:
1. Counts the workload node also when determining number of worker nodes (still excludes infra)
2. Exlcudes the workload node from backend pod placement using node anti-affinity

This means that while we still will have the same number of backend pods (2000), the workload node is ecluded from nginx pod placement, which should make results more deterministic as the workload node exclusively hosts the mb client

### DIFF
--- a/workloads/router-perf-v2/templates/nginx-deploy.yml
+++ b/workloads/router-perf-v2/templates/nginx-deploy.yml
@@ -15,7 +15,15 @@ spec:
       labels:
         app: nginx-{{.Replica}}
     spec:
-      nodeSelector: {{.nodeSelector}}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/workload
+                operator: DoesNotExist
+              - key: node-role.kubernetes.io/infra
+                operator: DoesNotExist
       containers:
       - name: nginx
         image: quay.io/cloud-bulldozer/nginx:latest
--- a/workloads/router-perf-v2/env.sh
+++ b/workloads/router-perf-v2/env.sh
@@ -7,7 +7,7 @@ export ES_SERVER=${ES_SERVER:-https://search-perfscale-dev-chmf5l4sh66lvxbnadi4b
 export ES_INDEX=${ES_INDEX:-router-test-results}
 
 # Environment setup
-NUM_NODES=$(oc get node -l node-role.kubernetes.io/worker,node-role.kubernetes.io/workload!=,node-role.kubernetes.io/infra!= --no-headers | grep -cw Ready)
+NUM_NODES=$(oc get node -l node-role.kubernetes.io/worker,node-role.kubernetes.io/infra!= --no-headers | grep -cw Ready)
 LARGE_SCALE_THRESHOLD=${LARGE_SCALE_THRESHOLD:-24}
 METADATA_COLLECTION=${METADATA_COLLECTION:-true}
 KUBE_BURNER_RELEASE_URL=${KUBE_BURNER_RELEASE_URL:-https://github.com/cloud-bulldozer/kube-burner/releases/download/v0.16.2/kube-burner-0.16.2-Linux-x86_64.tar.gz}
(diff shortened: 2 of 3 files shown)


rsevilla87:
If I understood correctly, the issue happens in managed-services based clusters where the workload node is labeled as worker too. If that's the case, I agree, to replace the nodeSelector by a nodeAffinity policy, however I don't understand why do you need to update the `NUM_NODES` expression, the original expression is not counting the workload node which is correct from my understanding, and given that no nginx pods would land to the workload node that will make it even more correct :)

rsevilla87:
> > If I understood correctly, the issue happens in managed-services based clusters where the workload node is labeled as worker too. If that's the case, I agree, to replace the nodeSelector by a nodeAffinity policy, however I don't understand why do you need to update the `NUM_NODES` expression, the original expression is not counting the workload node which is correct from my understanding, and given that no nginx pods would land to the workload node that will make it even more correct :)
> 
> In self-managed doesn't the worker node also get the workload label and same issue happens?
> 
> The reason i changed the expression to check for number of nodes is that we define the small scale/ large scale test based on number of worker nodes. We have traditionally been running tests on 24 node clusters and I want the large scale variant of the test to run there, but with this check, a 24 node cluster is counted as 23 nodes and the small scale test happens. Since multi-az clusters need to be a multiple of 3 my next option is a 27 node cluster - I cannot do 25 and that changes the control plane/infra sizing.
> 
> Tl;DR 24 is a standard size for pre-release testing and I would like to to see the large scale t4est run there as there is enough capacity and no reason not to.

Understood, so the scenario you're testing uses 24 worker nodes deployed across 3 AZs and one of those worker nodes is labeled as workload too. Have you considered creating a machinepool for the workload node?

workloads/router-perf-v2/templates/nginx-deploy.yml
```diff
@@ -15,7 +15,15 @@ spec:
       labels:
         app: nginx-{{.Replica}}
     spec:
-      nodeSelector: {{.nodeSelector}}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
```
chentex, line 20: This is new for me, it is quite specific

smalleni:
> > > If I understood correctly, the issue happens in managed-services based clusters where the workload node is labeled as worker too. If that's the case, I agree, to replace the nodeSelector by a nodeAffinity policy, however I don't understand why do you need to update the `NUM_NODES` expression, the original expression is not counting the workload node which is correct from my understanding, and given that no nginx pods would land to the workload node that will make it even more correct :)
> > 
> > 
> > In self-managed doesn't the worker node also get the workload label and same issue happens?
> > The reason i changed the expression to check for number of nodes is that we define the small scale/ large scale test based on number of worker nodes. We have traditionally been running tests on 24 node clusters and I want the large scale variant of the test to run there, but with this check, a 24 node cluster is counted as 23 nodes and the small scale test happens. Since multi-az clusters need to be a multiple of 3 my next option is a 27 node cluster - I cannot do 25 and that changes the control plane/infra sizing.
> > Tl;DR 24 is a standard size for pre-release testing and I would like to to see the large scale t4est run there as there is enough capacity and no reason not to.
> 
> Understood, so the scenario you're testing uses 24 worker nodes deployed across 3 AZs and one of those worker nodes is labeled as workload too. Have you considered creating a machinepool for the workload node?

Well I could, but the main point that is throwing me off is
1. We exclude the workload node from calculation of NUM_NODES and yet use it to launch the nginx pods
2. Most of our testing is at 24 nodes and when i spin up a 24 nodes I want to be able to just run the large version of the test without any additional steps

I do not see how 23 vs 24 will change the capacity materially for us to determine a large/small scale test, so unless someone can explain how that decision was made I think we should just use having 24 nodes overall in the cluster that have the worker label to determine the small/large version of the test.

rsevilla87 (approved):
lgtm

workloads/router-perf-v2/templates/nginx-deploy.yml
chentex: Yes, the affinity

PR #560, opened 2023-05-01
Removing iperf from the mix

jtaleric:
Why?
- Reduce the test time
- Having multiple load-drivers in CI might confuse the results.


jtaleric (commented):
going to self high-5 this one

PR #570, opened 2023-05-24
Index MC config metadata only when hypershift is detected

rsevilla87:
### Description

I think we can move this code to the hypershift() function


rsevilla87:
cc: @mukrishn

mukrishn (approved):
LGTM

PR #608, opened 2023-08-09
Deprecate legacy kube-burner workload

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Deprecation
- [ ] Documentation Update

## Description

The workload's on this directory shouldn't be used anymore, people should use https://github.com/cloud-bulldozer/e2e-benchmarking/tree/master/workloads/kube-burner-ocp-wrapper going forward.

### DIFF
--- a/ci/CI_test_list.yml
+++ b/ci/CI_test_list.yml
@@ -2,10 +2,4 @@ router-perf-v2:./ingress-performance.sh
 etcd-perf:./run_etcd_tests_fromgit.sh 
 scale-perf:./run_scale_fromgit.sh 
 network-perf:GEN_CSV=false ./run.sh
-kube-burner:WORKLOAD=cluster-density ./run.sh
-kube-burner:WORKLOAD=node-density ./run.sh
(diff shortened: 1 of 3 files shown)


rsevilla87:
Please, feel free to tag more affected users.

dry923:
@rsevilla87 Can we add an estimated removal date or something as well when the directory will be removed?

jtaleric:
Since this change only removes the CI runs I am good with it... I do think there are teams (ACS in particular) using non-ocp-wrapper kube-burner.

paigerube14:
Agree with Joe! Think it's ok to remove ci references. Not sure if we are fully ready (qe) side for a full removal of kube burner folder

rsevilla87:
> Agree with Joe! Think it's ok to remove ci references. Not sure if we are fully ready (qe) side for a full removal of kube burner folder

Thanks for the feedback!, as @dry923 also suggested, we should maybe add an EOL date so we can definitively remove the `kube-burner` directory

PR #628, opened 2023-09-26
Updated hypershift metric profiles to only read serving and non-serving nodes

mukrishn:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/workloads/kube-burner-ocp-wrapper/metrics-profiles/mc-metrics.yml
+++ b/workloads/kube-burner-ocp-wrapper/metrics-profiles/mc-metrics.yml
@@ -3,17 +3,12 @@
 # Only collecting container and worker nodes CPU/Memory metrics
 
 # Management Node metrics: CPU & Memory
+# Non-serving nodes
 - query: (sum(irate(node_cpu_seconds_total[1m])) by (mode,instance) and on (instance) label_replace(cluster:nodes_roles{label_hypershift_openshift_io_control_plane="true"}, "instance", "$1", "node", "(.+)")) > 0
   metricName: mgmt-nodeCPU-Workers
 
 - query: (node_memory_MemTotal_bytes - node_memory_MemAvailable_bytes) and on (instance) label_replace(cluster:nodes_roles{label_hypershift_openshift_io_control_plane="true"}, "instance", "$1", "node", "(.+)")
   metricName: mgmt-nodeMemoryUtilization-Workers
-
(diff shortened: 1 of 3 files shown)


krishvoor (approved):
/lgtm

workloads/kube-burner-ocp-wrapper/metrics-profiles/mc-metrics.yml
```diff
@@ -3,12 +3,18 @@
 # Only collecting container and worker nodes CPU/Memory metrics
 
 # Management Node metrics: CPU & Memory
-- query: (sum(irate(node_cpu_seconds_total[2m])) by (mode,instance) and on (instance) label_replace(kube_node_role{role="worker"}, "instance", "$1", "node", "(.+)")) > 0
+- query: (sum(irate(node_cpu_seconds_total[2m])) by (mode,instance) and on (instance) label_replace(cluster:nodes_roles{label_hypershift_openshift_io_control_plane="true"}, "instance", "$1", "node", "(.+)")) > 0
```
dry923, line 6: @mukrishn i believe this label only exists on the non-serving worker.s if we want the serving as well then we need hypershift.openshift.io/request-serving-component="true" too

mukrishn: True, this metric is to get the overall worker performance, so I intentionally want to remove the serving component because they are not shared and it won't affect over all performance(wrt 80HC capacity) and also it is per hc data so we cannot aggregate/average it other cluster. 
But to measure serving components, I've added two new metrics and that is going to be per hc - `mgmt-servingNode*`

mukrishn: And serving and non-serving might not be of same instance type, so it is better to keep them under different metric. 
Question is do we need to monitor default pool anymore ?

dry923 (approved):
LGTM!

mukrishn:
@rsevilla87 PTAL, I am reducing the rate interval from `2m` to `1m` to capture spike that happens periodically by some operator(catalog,registry) pods, aggressive on the nodes up to 4 cores for a few seconds. They get flatten with `2m` interval. please let me know if you think otherwise.

rsevilla87 (approved):
lgtm

PR #634, opened 2023-10-11
add ingress-perf config template to be overwirte

qiliRedHat:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/workloads/ingress-perf/run.sh
+++ b/workloads/ingress-perf/run.sh
@@ -15,6 +15,24 @@ TOLERANCY=${TOLERANCY:-20}
 OS=$(uname -s)
 HARDWARE=$(uname -m)
 
+export TUNING_PATCH=${TUNING_PATCH:-"'{\"spec\":{\"nodePlacement\": {\"nodeSelector\": {\"matchLabels\": {\"node-role.kubernetes.io/infra\": \"\"}}}, \"replicas\": 2}}'"}
+export CONNECTIONS=${CONNECTIONS:-200}
+export SAMPLES=${SAMPLES:-2}
+export DURATION=${DURATION:-5m}
+export ENDPOINT_PATH=${ENDPOINT_PATH:-'/1024.html'}
+export CONCURRENCY=${CONCURRENCY:-18}
+export TOOL=${TOOL:-wrk}
+export SERVER_REPLICAS=${SERVER_REPLICAS:-45}
+export REQUEST_TIMEOUT=${REQUEST_TIMEOUT:-10s}
+export DELAY=${DELAY:-10s}
+
+export CONFIG_TEMPLATE=${CONFIG_TEMPLATE}
+
+if [[ ! -z ${CONFIG_TEMPLATE}  ]]; then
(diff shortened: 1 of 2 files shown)


qiliRedHat:
I want to do two changes for ‘tuningPatch’ for a same test config, e.g.
oc -n openshift-ingress-operator patch ingresscontroller/default --type=merge -p '{"spec":{"tuningOptions": {"threadCount": 8}}}' 
oc -n openshift-ingress-operator patch ingresscontroller/default --type=merge -p '{"spec":{"nodePlacement": {"nodeSelector": {"matchLabels": {"node-role.kubernetes.io/infra": ""}}}, "replicas": 3}}'
I failed to put them together in one ‘tuningPatch’ https://github.com/cloud-bulldozer/ingress-perf/blob/main/pkg/runner/tuning.go#L35

qiliRedHat:
@rsevilla87 PTAL, in some of the tests such as more ingress replicas, I hope to be able to customize the parameters in the configuration files and avoid adding new configuration files for each non-standard tests configurations.

rsevilla87:
The change looks good technically, but I wonder, that if what you need new configurations, why don't you add more config files rather than templating the existing...
We're trying to avoid this kind of templates as much as possible as they are difficult to track.

qiliRedHat:
> The change looks good technically, but I wonder, that if what you need new configurations, why don't you add more config files rather than templating the existing... We're trying to avoid this kind of templates as much as possible as they are difficult to track. Same applies to other tools

@rsevilla87 Thanks for the comment, I'll consider to use new configurations. I can overwrite config as env var so no need to use custom-run.sh.

Please let me know if you have some suggestion for https://github.com/cloud-bulldozer/e2e-benchmarking/pull/634#issuecomment-1757794801

workloads/ingress-perf/run.sh
```diff
@@ -18,6 +18,24 @@ TOLERANCY=${TOLERANCY:-20}
 OS=$(uname -s)
 HARDWARE=$(uname -m)
 
+export TUNING_PATCH=${TUNING_PATCH:-"'{\"spec\":{\"nodePlacement\": {\"nodeSelector\": {\"matchLabels\": {\"node-role.kubernetes.io/infra\": \"\"}}}, \"replicas\": 2}}'"}
```
rsevilla87, line 21: Rather than modifying `run.sh`, can you push this changes into a new file  `custom-run.sh`? that way we won't touch any of the current logic in CPT

qiliRedHat: I have used new configurations to run different replica numbers. I find t is more easier than custom-run.sh for me that I can just provide the new config file as an env var, no need to update the CI to use a new cmd like custom-run.sh. So I think I can close this one. 

@rsevilla87  If you have any suggestion to
https://github.com/cloud-bulldozer/e2e-benchmarking/pull/634#issuecomment-1757794801, it could be helpful. Now to tune the thread number, I do it outside of ingress-perf with oc command before the ingress-perf test.

PR #638, opened 2023-10-16
Use workload script to call kube-burner index

mukrishn:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
For hypershift we use kube-burner index to scrape metrics during creation/deletions of clusters to record the MC perf, so instead of re-implementing these(multi-endpoint prep) in rosa-burner it would be easy to use the same script with a conditional tasks. 

WORKLOAD=index --> will only call kube-burner index for the given duration START_TIME & END_TIME

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/workloads/kube-burner-ocp-wrapper/run.sh
+++ b/workloads/kube-burner-ocp-wrapper/run.sh
@@ -98,7 +98,7 @@ EOF
 
 download_binary
 if [[ ${WORKLOAD} =~ "index" ]]; then
-  cmd="${KUBE_DIR}/kube-burner index --uuid=${UUID} --start=$START_TIME --end=$((END_TIME+600)) --log-level debug"
+  cmd="${KUBE_DIR}/kube-burner index --uuid=${UUID} --start=$START_TIME --end=$((END_TIME+600)) --log-level ${LOG_LEVEL}"
 else
   cmd="${KUBE_DIR}/kube-burner ocp ${WORKLOAD} --log-level=${LOG_LEVEL} --qps=${QPS} --burst=${BURST} --gc=${GC} --uuid ${UUID}"
   cmd+=" ${EXTRA_FLAGS}"
--- a/workloads/kube-burner-ocp-wrapper/run.sh
+++ b/workloads/kube-burner-ocp-wrapper/run.sh
@@ -78,7 +78,7 @@ EOF
 
   if [[ ${WORKLOAD} =~ "index" ]]; then
     cat << EOF
-elapsed: "20m:"
+elapsed: "20m"
 EOF
   fi
   
--- a/workloads/kube-burner-ocp-wrapper/run.sh
+++ b/workloads/kube-burner-ocp-wrapper/run.sh
@@ -75,6 +75,13 @@ HOSTED_PROMETHEUS_TOKEN: <truncated>
 HCP_NAMESPACE: ${HCP_NAMESPACE}
 MGMT_WORKER_NODES: ${MGMT_WORKER_NODES}
 EOF
+
+  if [[ ${WORKLOAD} =~ "index" ]]; then
+    cat << EOF
+elapsed: "20m:"
+EOF
+  fi
+  
   echo "Indexing Management cluster stats before executing"
   METADATA=$(cat << EOF
 {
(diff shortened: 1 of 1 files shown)


vishnuchalla:
@mukrishn I think this change is not required as we plan to release index sub-command  as a part of OCP wrapper itself: https://github.com/cloud-bulldozer/kube-burner/pull/451

mukrishn:
@vishnuchalla IIUC the PR, it is just to enable index option in ocp wrapper,  even then I will use `index` sub-command to scrape without any workload.

workloads/kube-burner-ocp-wrapper/run.sh
```diff
@@ -75,6 +75,13 @@ HOSTED_PROMETHEUS_TOKEN: <truncated>
 HCP_NAMESPACE: ${HCP_NAMESPACE}
 MGMT_WORKER_NODES: ${MGMT_WORKER_NODES}
 EOF
+
+  if [[ ${WORKLOAD} =~ "index" ]]; then
+    cat << EOF
+elapsed: "20m:"
+EOF
+  fi
+
```
rsevilla87, line 84: What's this for?

workloads/kube-burner-ocp-wrapper/run.sh
```diff
@@ -90,7 +97,12 @@ EOF
 }
 
 download_binary
-cmd="${KUBE_DIR}/kube-burner ocp ${WORKLOAD} --log-level=${LOG_LEVEL} --qps=${QPS} --burst=${BURST} --gc=${GC} --uuid ${UUID}"
+if [[ ${WORKLOAD} =~ "index" ]]; then
+  cmd="${KUBE_DIR}/kube-burner index --uuid=${UUID} --start=$START_TIME --end=$((END_TIME+600)) --log-level debug"
```
rsevilla87, line 101:
```suggestion
  cmd="${KUBE_DIR}/kube-burner index --uuid=${UUID} --start=$START_TIME --end=$((END_TIME+600)) --log-level ${LOG_LEVEL}"
```

rsevilla87:
@mukrishn does this PR need to use the metrics-endpoint flag?,

I think not, from the PR description, I assume you're trying to collect metrics just from the management cluster, if that's correct. You can use `kube-burner ocp index` as @vishnuchalla suggested, and rather than setting `MC_KUBECONFIG` you can just point the regular `KUBECONFIG` variable to the mngt-cluster's kubeconfig

workloads/kube-burner-ocp-wrapper/run.sh
mukrishn: kube-burner set this variable at workload execution to limit the duration of certain metrics like [this](https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/kube-burner-ocp-wrapper/metrics-profiles/mc-metrics.yml#L16) so I am hard coding here to 20m. Yes I should remove the `:`.

mukrishn:
@rsevilla87 yes I can use, `kube-burner ocp index` with `--metric-profile` pointed to my MC profile, but is this failure a limitation or it is happening only on MC?  
I can work around by avoiding `--metric-endpoint` in this case, but I thought it would be useful to support this if I want to collect metrics without workload.

workloads/kube-burner-ocp-wrapper/run.sh
rsevilla87: why 20m? what's wrong with the original .elapsed? if you don't need it anymore, set everything to 20m to avoid confusion

mukrishn: I guess there is confusion here, idea is to call only index without workload so I assume elapsed will not be set by kube-burner based on [this](https://github.com/cloud-bulldozer/kube-burner/blob/71a4ba5ec036650cf9827235746dd84cbf73cb74/docs/observability/metrics.md?plain=1#L73). there is no workload executed by kube-burner at first place to get the duration of job. 
If my understanding is wrong, I can remove this elapsed variable from here.

rsevilla87: oh, that's right, elapsed won't be set unless your using this profile from a benchmark. Now I wonder...
These lines are correct?
```shell
  if [[ ${WORKLOAD} =~ "index" ]]; then
    cat << EOF
elapsed: "20m"
EOF
  fi
```

That statement print `elapsed: "20"` in the standard output, shouldn't be `export elapsed=20m`?

mukrishn: right, replaced it.

rsevilla87 (approved):
lgtm!

PR #652, opened 2023-11-15
New way of comparing current run uuid with matching data runs

paigerube14:
## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
With an environment variable set of the current UUID of a run, we want to be able to see if any metrics have regressed enough to fail a job. This current implementation takes the metadata of the uuid and collects other matching runs. Using the similar runs it aggregates the data of specific fields and gives a pass fail based on toleration value 

This is still a work in progress and would love any feedback. I tried to document steps and configurations in the utils/compare/README.md 

- Related Issue #https://issues.redhat.com/browse/PERFSCALE-2298

- [X] I have performed a self-review of my code.

## Testing

Example ingress comparison
```
uuid - value_y: 833d28b4-23ad-481a-bbce-7d0735f14fef
comparison ids - value_x: ['perfscale-cpt-84173e88-bbe6-4e97-9a88-cde0b12b693b', 'f5544cff-b022-43b8-8918-965d987b6c44', 'CPT-d2581a63-b46b-48bc-a298-53cfee346bf7', 'e9cb2f03-4bdd-468d-9940-6dd7150f2cf3', '6297dffe-0987-4614-8858-2806926571a4', 'fda3af38-7181-4d03-ac38-e32a7d7e1dfb', '80ec50dd-6ef8-4b76-a614-3525cae5ebac', '96c6b369-cf90-409f-ba51-5f852665d89a']
,config.termination,sample,config.serverReplicas,config.connections,config.concurrency,total_avg_rps_x,total_avg_rps_y,difference,result
0,edge,1,45,200,18,91824.4242857143,93677.84999999999,2.0184452325219038,Pass
1,edge,2,45,200,18,92859.28428571428,97639.56999999998,5.147881281937794,Pass
2,http,1,45,200,18,137532.44142857144,143860.77000000002,4.6013351509616385,Pass
3,http,2,45,200,18,138334.88571428572,144228.68000000002,4.260526370685147,Pass
4,passthrough,1,45,200,18,212393.5,232025.31000000003,9.243131263433213,Pass
5,passthrough,2,45,200,18,215646.17142857146,261904.77000000002,21.45115689510435,Fail
6,reencrypt,1,45,200,18,77213.55428571429,82293.16,6.578645111309855,Pass
7,reencrypt,2,45,200,18,76440.35285714285,82400.04000000001,7.79651966546393,Pass
```


vishnuchalla:
Can we `gitignore` the `__pycache__` please?

utils/compare/README.md
```diff
@@ -0,0 +1,99 @@
...
+* configs/crio.json
+* configs/kubelet.json
+* configs/etcd.json
+* configs/master_node.json
+* configs/worker_node.json*
+
+## Defining new config
+
+Each config is a json file consisting of 
+* metrics: A list of dictionairies each contiaining a list of the key values you want to compare upon (can use regex on some values, depends on set up in elasticsearch). Multiple items in this list will still use the same below fields with every search in addition to the key/values set in the dictionary
```
vishnuchalla, line 84: Can we specify those `same below fields` used in addition to the key/values set in the dictionary, please?

utils/compare/README.md
```diff
@@ -0,0 +1,99 @@
...
+* additionalColumns: List of keys that you want to have in your results table, this can be used for only combining certain metrics. For example. if I set "samples" in the additionalColumns the data wont combine/aggregate different samples, it'll give me separate rows for each different sample type
```
vishnuchalla, line 85: Can we specify those certain metrics that are `combined/not-combined` for easy understanding.

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+    return runs
+
+def getMatchRuns(meta: dict, workerCount: False, previous_version=False):
+    index = "perf_scale_ci"
+    version = meta["ocpVersion"][:4]
+    if previous_version: 
+        version = float(version) - .01
+    
+    # https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html
+    if not workerCount:
```
vishnuchalla, line 230:
```suggestion
        query_params = {
        "benchmark": meta["benchmark"],
        "workerNodesType": meta["workerNodesType"],
        "masterNodesType": meta["masterNodesType"],
        "networkType": meta["networkType"],
        "platform": meta["platform"],
        "ocpVersion": f"{version}*",
        "jobStatus": "success",
    }

    if workerCount:
        query_params.update({
            "masterNodesCount": meta["masterNodesCount"],
            "workerNodesCount": meta["workerNodesCount"],
        })

    query = {
        "_source": ["uuid"],
        "query": {
            "query_string": {
                "query": " AND ".join([f'{key}: "{value}"' for key, value in query_params.items()]),
            }
        }
    }

    es = ElasticService(index=index)
    response = es.post(query)
    es.close()
    return [item['_source']['uuid'] for item in response["hits"]["hits"]]
```
This will only fetch the `uuid` portion of the docs instead of the entire docs in-memory.

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+        'type' : 'bar',
+        'orientation' : 'v'}
+    x=[]
+    y=[]
+    for index, row in nMetrics.iterrows():
+        test = "{}-{}".format(row['profile'], row['messageSize'])
+        value = "{}".format(row['throughput'])
+        x.append(value)
+        y.append(test)
+    new = {'y' : x,
```
vishnuchalla, line 29: These `x, y` need not necessarily interchange. This implementation can be written in more readable and simple way. More details [here](https://github.com/cloud-bulldozer/cpt-dashboard/blob/1146c0e0f2506e5b5cfd8ea37e379519fa4c0e96/backend/app/api/v1/endpoints/quay/quayGraphs.py#L36-L97).

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+def pd_merge(previous_data, current_data, columns):
+    pd_merged = current_data.merge(previous_data, how='outer', on=columns[:-1])
+    return pd_merged
+
+def find_result(row, tolerancy): 
+
+    if np.isinf(row['difference']): 
+        return "Pass"
+    # Fail if value is a more negative than tolerancy
+    # High positive valus mean new uuid is better
```
vishnuchalla, line 84: [typo]: `High positive values`

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+
+def pd_merge(previous_data, current_data, columns):
+    pd_merged = current_data.merge(previous_data, how='outer', on=columns[:-1])
+    return pd_merged
+
+def find_result(row, tolerancy): 
+
+    if np.isinf(row['difference']): 
+        return "Pass"
+    # Fail if value is a more negative than tolerancy
```
vishnuchalla, line 83: `README.md` line#16 says better performance in negative case, but here we are marking it as Fail 🤔

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+        'name' : 'Current results average',
+        'type' : 'bar',
+        'orientation' : 'v'}
+    metrics.append(old)
+    metrics.append(new)
+
+    return metrics
+
+
+def get_final_graph_metrics(metric_of_interest, divider,previous_data, current_data):
```
vishnuchalla, line 40: What is the idea behind these plotly graph functions? I couldn't see any invocations of them in this PR.

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+def burnerFilter(data: dict, columns: list[str]) :
+    #
+    # Filter out aspects of the test to norm results
+    #
+    #pprint.pprint(data)
+    ndf = pd.DataFrame(data, columns=columns)
+    print("ndf " + str(ndf))
+    return ndf
+
+def getResults(uuid: str, uuids: list, index_str: str, metrics: dict ):
```
vishnuchalla, line 193: The idea behind this function looks cool. I think we can reuse the same to fetch other metrics like `jobSummary` etc.

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+    #pprint.pprint(data)
+    df = pd.json_normalize(data)
+    filterDF = burnerFilter(df, columns)
+    if func == "avg" or func == "mean": 
+        ptile = filterDF.groupby(columns[:-1])[columns[-1]].mean()
+    elif func == "max": 
+        ptile = filterDF.groupby(columns[:-1])[columns[-1]].max()
+    return ptile
+
+def processBurner(data: dict) :
```
vishnuchalla, line 165: Is this function being used somewhere?

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+    #pprint.pprint(data)
+    df = pd.a(data)
+    filterDF = burnerFilter(df)
+    #print('filter df ' + str(filterDF))
+    ptile = filterDF.groupby(['quantileName'])['P99'].mean()
+    #print("p tile" + str(ptile))
+    return ptile
+
+def jobFilter(pdata: dict, data: dict):
```
vishnuchalla, line 174: No invocations for this function call too. Let remove dead code as it creates unnecessary confusion.

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+def generate_pass_fail(oMetrics, nMetrics, columns, tolerancy_num):
+
+    all_merged = pd_merge(oMetrics, nMetrics, columns)
+    print('all merged' + str(all_merged))
+    if not all_merged.empty:
+        all_result_tolerancy = tolerancy(all_merged, tolerancy_num)
+        return all_result_tolerancy
+    return []
+
+def jobSummary(uuids: list):
```
vishnuchalla, line 135: just now realized that, this function can also be removed.

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+    pdf = pd.json_normalize(pdata)
+    pick_df = pd.DataFrame(pdf, columns=columns)
+    iterations = pick_df.iloc[0]['jobConfig.jobIterations']
+    df = pd.json_normalize(data)
+    ndf = pd.DataFrame(df, columns=columns)
+    ids_df = ndf.loc[df['jobConfig.jobIterations'] == iterations ]
+    return ids_df['uuid'].to_list()
+
+def burnerFilter(data: dict, columns: list[str]) :
+    #
```
vishnuchalla, line 185: If interested in keeping doc strings, we can keep the format same across: https://google.github.io/styleguide/pyguide.html

utils/compare/read_files.py
```diff
@@ -0,0 +1,104 @@
...
+    uuid_data.append(id_data)
+    print('uuid dat ' + str(uuid_data))
+    with open(filepath, 'w') as f:
+     
+        # using csv.writer method from CSV package
+        write = csv.writer(f)
+        write.writerows(uuid_data)
+
+data_func = "mean"
+previous_version = os.getenv("previous_version", False)
```
vishnuchalla, line 30: can we keep all the envs `CAPITALIZED` please.

utils/compare/read_files.py
```diff
@@ -0,0 +1,104 @@
...
+    metrics = [ingress_json]
+else:
+    index = "ripsaw-kube-burner"
+
+    latency_json = read_config_file("configs/podlatency.json")
+    crio_json = read_config_file("configs/crio.json")
+    kubelet_json = read_config_file("configs/kubelet.json")
+    etcd_json = read_config_file("configs/etcd.json")
+    master_json = read_config_file("configs/master_node.json")
+    worker_agg_json = read_config_file("configs/worker_node.json")
```
vishnuchalla, line 56: Did we miss on adding/removing this from the metrics list?

utils/compare/search.py
```diff
@@ -0,0 +1,34 @@
+from elasticsearch import Elasticsearch
+from fastapi.encoders import jsonable_encoder
+import os
+# elasticsearch constants
+ES_URL = 'search-ocp-qe-perf-scale-test-elk-hcm7wtsqpxy7xogbu72bor4uve.us-east-1.es.amazonaws.com'
```
vishnuchalla, line 5: Can we keep this as ENV too? Just don't wanted to get it exposed to the outside world.

vishnuchalla:
Do we want to remove or name first column? Currently it seems empty
```
,metricName,quantileName,P99_x,P99_y,difference,result
0,podLatencyQuantilesMeasurement,ContainersReady,13.0,13.0,0.0,Pass
1,podLatencyQuantilesMeasurement,Initialized,1.0,1.0,0.0,Pass
```

vishnuchalla (changes requested):
In overall I like this idea. After going through the discussion on JIRA attached, _Reg: "I'm wondering where a good spot more helper type functions should go"_, here are my .02 cents
1.  If we want visualizations on top of this data, `cpt-dashboard` would be a better place to go.
2. Or else if we just want to use it as a tool at the end of each CI run just to look at the results in a `csv/tabular/json` format, I think better place would be `benchmark-comparison` that we have as of today.

utils/compare/graph.py
```diff
@@ -0,0 +1,306 @@
...
+    all_with_result = all_data_points.assign(difference=result)
+
+    all_with_result['result'] = all_with_result.apply(lambda row: find_result(row, tolerancy_percent), axis=1)
+
+    return all_with_result
+
+def process_all_data(metric_of_interest, find_metrics,uuid, ids, data_func, index, divider, additional_columns): 
+    # Process all
+
+    columns = []
```
vishnuchalla, line 106: Can we make it a `set` please, so that we can avoid duplicate columns.

utils/compare/graph.py
jtaleric: I think a lot of these are from the graph-api in the dashboard work... 

Preferably, we should create a python package to share these functions, and reuse them across our projects.

paigerube14:
> Do we want to remove or name first column? Currently it seems empty
> 
> ```
> ,metricName,quantileName,P99_x,P99_y,difference,result
> 0,podLatencyQuantilesMeasurement,ContainersReady,13.0,13.0,0.0,Pass
> 1,podLatencyQuantilesMeasurement,Initialized,1.0,1.0,0.0,Pass
> ```

I think this is just the row indices from the data frame, I will look into how I can easily remove that

utils/compare/graph.py
paigerube14: it was an old function I rewrote to work for all workload types, should be taken out now

utils/compare/graph.py
paigerube14: it was an old function I rewrote to work for all workload types, should be taken out now

utils/compare/graph.py
vishnuchalla: Expect for the `cpt-dashboard` what is the alternative place that we are planning to visualize these graphs?

utils/compare/search.py
paigerube14: took these variables out and just using the ES_SERVER variable like the rest of kube-burner/other workloads export

utils/compare/graph.py
jtaleric: Honestly, no where to my knowledge. The idea was to share how we build the dataframes for the trending graphs and for the comparisons.

utils/compare/graph.py
paigerube14: I had just switched the way the dataframe was set up switching the percent values, will update to be sure they match

utils/compare/read_files.py
paigerube14: Yes! Think I updated now

utils/compare/graph.py
paigerube14: Because I use this variable as the columns for the dataframe I dont think it allows me to specify it as a set. 


  ```
File "/Users/prubenda/PycharmProjects/e2e-benchmarking/utils/compare/venv3/lib/python3.10/site-packages/pandas/core/frame.py", line 702, in __init__
    raise ValueError("columns cannot be a set")
ValueError: columns cannot be a set
```

utils/compare/graph.py
vishnuchalla: IMHO, lets wait for at-least on more tool to have the same use-case of displaying those graphs. Until then its just dead code and can be removed.

utils/compare/graph.py
vishnuchalla: In that case, we can have set (to handle duplicates if any in the additionalColumns) to collect all the columns and then type cast it to a list which is supported by data frames.

jtaleric:
> In overall I like this idea. After going through the discussion on JIRA attached, _Reg: "I'm wondering where a good spot more helper type functions should go"_, here are my .02 cents
> 
> 1. If we want visualizations on top of this data, `cpt-dashboard` would be a better place to go.

IMHO the shared functions should go into a shared library -- I am thinking something similar to `go-commons` -- `python-commons`? 

> 2. Or else if we just want to use it as a tool at the end of each CI run just to look at the results in a `csv/tabular/json` format, I think better place would be `benchmark-comparison` that we have as of today.

The implementation belongs in benchmark-comparison -- agreed. Or we create a tool more specific to the use case `benchmark-regression`

jtaleric:
> > In overall I like this idea. After going through the discussion on JIRA attached, _Reg: "I'm wondering where a good spot more helper type functions should go"_, here are my .02 cents
> > 
> > 1. If we want visualizations on top of this data, `cpt-dashboard` would be a better place to go.
> 
> IMHO the shared functions should go into a shared library -- I am thinking something similar to `go-commons` -- `python-commons`?
> 
> > 2. Or else if we just want to use it as a tool at the end of each CI run just to look at the results in a `csv/tabular/json` format, I think better place would be `benchmark-comparison` that we have as of today.
> 
> The implementation belongs in benchmark-comparison -- agreed. Or we create a tool more specific to the use case `benchmark-regression`

> In overall I like this idea. After going through the discussion on JIRA attached, _Reg: "I'm wondering where a good spot more helper type functions should go"_, here are my .02 cents
> 
> 1. If we want visualizations on top of this data, `cpt-dashboard` would be a better place to go.

IMHO the shared functions should go into a shared library -- I am thinking something similar to `go-commons` -- `python-commons`?

We have started - 
https://github.com/cloud-bulldozer/py-commons


> 2. Or else if we just want to use it as a tool at the end of each CI run just to look at the results in a `csv/tabular/json` format, I think better place would be `benchmark-comparison` that we have as of today.

The implementation belongs in benchmark-comparison -- agreed. Or we create a tool more specific to the use case `benchmark-regression`

> > In overall I like this idea. After going through the discussion on JIRA attached, _Reg: "I'm wondering where a good spot more helper type functions should go"_, here are my .02 cents
> > 
> > 1. If we want visualizations on top of this data, `cpt-dashboard` would be a better place to go.
> 
> IMHO the shared functions should go into a shared library -- I am thinking something similar to `go-commons` -- `python-commons`?
> 
> > 2. Or else if we just want to use it as a tool at the end of each CI run just to look at the results in a `csv/tabular/json` format, I think better place would be `benchmark-comparison` that we have as of today.
> 
> The implementation belongs in benchmark-comparison -- agreed. Or we create a tool more specific to the use case `benchmark-regression`

> In overall I like this idea. After going through the discussion on JIRA attached, _Reg: "I'm wondering where a good spot more helper type functions should go"_, here are my .02 cents
> 
> 1. If we want visualizations on top of this data, `cpt-dashboard` would be a better place to go.

IMHO the shared functions should go into a shared library -- I am thinking something similar to `go-commons` -- `python-commons`?

> 2. Or else if we just want to use it as a tool at the end of each CI run just to look at the results in a `csv/tabular/json` format, I think better place would be `benchmark-comparison` that we have as of today.

The implementation belongs in benchmark-comparison -- agreed. Or we create a tool more specific to the use case `benchmark-regression`

> In overall I like this idea. After going through the discussion on JIRA attached, _Reg: "I'm wondering where a good spot more helper type functions should go"_, here are my .02 cents
> 
> 1. If we want visualizations on top of this data, `cpt-dashboard` would be a better place to go.

IMHO the shared functions should go into a shared library -- I am thinking something similar to `go-commons` -- `python-commons`?

> 2. Or else if we just want to use it as a tool at the end of each CI run just to look at the results in a `csv/tabular/json` format, I think better place would be `benchmark-comparison` that we have as of today.

The implementation belongs in benchmark-comparison -- agreed. Or we create a tool more specific to the use case `benchmark-regression`

> > In overall I like this idea. After going through the discussion on JIRA attached, _Reg: "I'm wondering where a good spot more helper type functions should go"_, here are my .02 cents
> > 
> > 1. If we want visualizations on top of this data, `cpt-dashboard` would be a better place to go.
> 
> IMHO the shared functions should go into a shared library -- I am thinking something similar to `go-commons` -- `python-commons`?
> 
> > 2. Or else if we just want to use it as a tool at the end of each CI run just to look at the results in a `csv/tabular/json` format, I think better place would be `benchmark-comparison` that we have as of today.
> 
> The implementation belongs in benchmark-comparison -- agreed. Or we create a tool more specific to the use case `benchmark-regression`

jtaleric:
> > In overall I like this idea. After going through the discussion on JIRA attached, _Reg: "I'm wondering where a good spot more helper type functions should go"_, here are my .02 cents
> > 
> > 1. If we want visualizations on top of this data, `cpt-dashboard` would be a better place to go.
> 
> IMHO the shared functions should go into a shared library -- I am thinking something similar to `go-commons` -- `python-commons`?
https://github.com/cloud-bulldozer/py-commons

We have created this repo ;) 

> 
> > 2. Or else if we just want to use it as a tool at the end of each CI run just to look at the results in a `csv/tabular/json` format, I think better place would be `benchmark-comparison` that we have as of today.
> 
> The implementation belongs in benchmark-comparison -- agreed. Or we create a tool more specific to the use case `benchmark-regression`

paigerube14:
Moving this work between py-commons and orion

A start of moving this functionality is in these PR's 

https://github.com/cloud-bulldozer/orion/pull/9
https://github.com/cloud-bulldozer/py-commons/pull/16/

PR #653, opened 2023-11-23
Decrease concurrency to 5 on router tests for rosa

morenod:
Because of the size of the infra nodes when installing rosa, we need to decrease this parameter to 5


rsevilla87:
ADR of the change https://docs.google.com/document/d/1HvR1JWLD4OncO8db-DXCjQ2qbD96dJGkbMP3mUq7qdM/edit

rsevilla87 (approved):
lgtm

PR #654, opened 2023-11-27
moved api latency queries to guest side

mukrishn:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description
API latency metrics scrapped by OBO are only for SRE alerting purpose, they are limited to collect only few buckets to save storage due to its volume.  But all KAPI metrics are forwarded to guest cluster side as well for user benefits, we should be able to scrape them from hosted cluster prometheus instance, so moving API latency metrics to hosted-cluster-metric profile will be the right approach for our reports.

- Related Issue #
- Closes #

## Testing


rsevilla87 (approved):
lgtm!

PR #672, opened 2024-02-07
PERFSCALE-2825: Add curl retry flags

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When  curl  is  about to retry a transfer, it will first wait one second and then for all forthcoming retries it will double the waiting time until it reaches 10 minutes...
Regarding the `--retry-all-errors` flag, taken from the cURL docs

> By  default  curl will not error on an HTTP response code that indicates an HTTP error, if the transfer was successful. For example, if a server replies 404 Not Found and the reply is fully received then that is not an error. When --retry is used then curl will retry on some HTTP response codes that indicate transient HTTP errors, but that does not include most 4xx response codes such as 404. If you want  to  retry  on  all  response codes that indicate HTTP errors (4xx and 5xx) then combine with -f, --fail.


mohit-sheth (approved):
lgtm

vishnuchalla (approved):
Strange to see that this issue is still present. LGTM.

jtaleric (approved):
lgtm!

PR #682, opened 2024-02-20
Updates IPSEC method for 4.15

chentex:
## Type of change

- [x] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Updates IPSEC method for 4.15
This should also avoid the errors that have been appearing in older versions

## Related Tickets & Documents

- Related Issue #
- Closes #

## Checklist before requesting a review

- [ ] I have performed a self-review of my code.
- [ ] If it is a core feature, I have added thorough tests.

## Testing
- Please describe the System Under Test.
- Please provide detailed steps to perform tests related to this code change.
- How were the fix/results from this change verified? Please provide relevant screenshots or results.

PR #684, opened 2024-02-21
WIP- Add whereabouts

mrbojangles3:
## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This directory contains the files to create and run a node-pod-density test with a net-attach-def applied to each pod. The goal is to test the whereabouts Ipam. This is a pretty heavy copy/paste from the existing node-density test, with changes for whereabouts. The is a separate directory so that the kube-burner directory is clean. Happy to move it there if people agree that is the right path forward. 

- Related Issue # [PERFSCALE-2680](https://issues.redhat.com//browse/PERFSCALE-2680)

- [x] I have performed a self-review of my code. it has also done some small runs in AWS.

## Testing


jtaleric:
> we can probably have this in kube-burner-ocp-wrapper maybe we need to remove "ocp" from it or have a common dir for metrics/alerts profiles

I wouldn't be opposed to this or keeping it within its own workloads directory... The blast radius of the change is contained with the current approach...

vishnuchalla:
> /test e2e-cluster-density-v2-single-node
just a test

jtaleric:
@vishnuchalla any diagnostic on why we are seeing the git message?

vishnuchalla:
Yes. It requires an update in the fork with latest mainstream changes and then a merge of it in the PR's branch. It resolved the issue for Vicente's [PR](https://github.com/cloud-bulldozer/e2e-benchmarking/pull/682#issuecomment-1966848167).

jtaleric:
> Yes. It requires an update in the fork with latest mainstream changes and then a merge of it in the PR's branch. It resolved the issue for Vicente's [PR](https://github.com/cloud-bulldozer/e2e-benchmarking/pull/682#issuecomment-1966848167).

@mrbojangles3 can you please update?

workloads/whereabouts/metrics-profiles/mc-metrics.yml
```diff
@@ -0,0 +1,96 @@
+# Management cluster metrics
```
jtaleric, line 1: Do we care to have this?

workloads/whereabouts/metrics-profiles/hosted-cp-metrics.yml
```diff
@@ -0,0 +1,32 @@
+# Hosted control-plane metrics
```
jtaleric, line 1: Do we care to have this? Not really testing HCP w/ Multus as of now, not sure we need to keep this.

workloads/whereabouts/metrics-profiles/hosted-cluster-metrics.yml
```diff
@@ -0,0 +1,106 @@
+# Hosted cluster metrics
```
jtaleric, line 1: Do we care to have this? Not testing multus w/ Hosted Clusters, right?

workloads/whereabouts/alerts-profiles/hosted-cp-alerts.yml
```diff
@@ -0,0 +1,36 @@
+# etcd
```
jtaleric, line 1: Should we discard this?

workloads/whereabouts/metrics-report.yml
```diff
@@ -0,0 +1,259 @@
...
+
+- query: avg(avg_over_time(topk(3,sum(container_memory_rss{name!="", namespace="openshift-etcd"}) by (pod))[{{.elapsed}}:]))
+  metricName: memory-etcd
+  instant: true
+
+- query: max(max_over_time(topk(3,sum(container_memory_rss{name!="", namespace="openshift-etcd"}) by (pod))[{{.elapsed}}:]))
+  metricName: max-memory-etcd
+  instant: true
+
+ # multus
```
jtaleric, line 110: should we add a query for the other components of multus? ie whereabouts?

workloads/whereabouts/node-density.yml
```diff
@@ -0,0 +1,36 @@
...
+    esServers: ["{{.ES_SERVER}}"]
+    insecureSkipVerify: true
+    defaultIndex: {{.ES_INDEX}}
+    type: opensearch
+  measurements:
+    - name: podLatency
+      thresholds:
+        - conditionType: Ready
+          metric: P99
+          threshold: 2000ms
```
jtaleric, line 15: We know multus cannot achieve this, should we consider higher threshold here?

workloads/whereabouts/run.sh
```diff
@@ -0,0 +1,164 @@
...
+set -e
+
+ES_SERVER=${ES_SERVER=https://search-perfscale-dev-chmf5l4sh66lvxbnadi4bznl3a.us-west-2.es.amazonaws.com}
+LOG_LEVEL=${LOG_LEVEL:-info}
+if [ "$KUBE_BURNER_VERSION" = "default" ]; then
+    unset KUBE_BURNER_VERSION
+fi
+KUBE_BURNER_VERSION=${KUBE_BURNER_VERSION:-1.1.0}
+CHURN=${CHURN:-false}
+WORKLOAD='node-density'
```
jtaleric, line 12: This should be something more specific like `whereabouts-node-density`

workloads/whereabouts/run.sh
```diff
@@ -0,0 +1,164 @@
...
+download_binary(){
+  KUBE_BURNER_URL="https://github.com/kube-burner/kube-burner-ocp/releases/download/v${KUBE_BURNER_VERSION}/kube-burner-ocp-V${KUBE_BURNER_VERSION}-linux-x86_64.tar.gz"
+  curl --fail --retry 8 --retry-all-errors -sS -L "${KUBE_BURNER_URL}" | tar -xzC "${KUBE_DIR}/" kube-burner-ocp
+}
+function cleanup_whereabouts(){
+    
+    # remove IP pool
+    oc delete ippools.whereabouts.cni.cncf.io '10.1.0.0-21' -n openshift-multus
+
+    overlaps=$(oc get overlappingrangeipreservations.whereabouts.cni.cncf.io -n openshift-multus --no-headers=true | awk '{print $1}');
```
jtaleric, line 30: shouldn't this be part of the exit criteria? After the workload completes and cleans up -- ie kube-burner deletes all the pods that it created, we should not have any IP addresses... if we do, we should exit 1 and have the status be that we leaked IPs, right?

workloads/whereabouts/run.sh
```diff
@@ -0,0 +1,164 @@
...
+        exit $exit_code
+    fi
+    # give the pods time to come up.
+    sleep_time=10
+    echo "sleeping for $sleep_time seconds to allow reconciler pods to start"
+    sleep $sleep_time
+}
+
+
+hypershift(){
```
jtaleric, line 60: Are we doing anything with hypershift -- i realize you might have just taken the existing `run.sh` but lets clean it up to be more specific for this work.

jtaleric (changes requested):
Some feedback thus far! Thanks!

mrbojangles3:
w.r.t the code that is brought in from the copy/paste. I was thinking that we might need it in the future and we would rather having hanging around unused vs bringing it all in when we did need it. I am assuming that those areas won't change much but I could be wrong? @jtaleric thoughts? I am good either way just figured I would explain myself.

jtaleric:
> w.r.t the code that is brought in from the copy/paste. I was thinking that we might need it in the future and we would rather having hanging around unused vs bringing it all in when we did need it. I am assuming that those areas won't change much but I could be wrong? @jtaleric thoughts? I am good either way just figured I would explain myself.

With HC I am not sure we have access to modify the deployment for whereabouts -- I just don't want there to be assumptions 6-12 months from now if we leave these vars in place and they do nothing.

workloads/whereabouts/node-density.yml
mrbojangles3: Good point, how much do we want to allow for "multus overhead"? I would be okay with adding a whole second for it. Do we want to throw a question toward the multus team?

workloads/whereabouts/run.sh
mrbojangles3: This is passed straight to kube-burner-ocp, so it would fail since there is no workload with that name. I was keeping it as is for simplicity.

workloads/whereabouts/alerts-profiles/hosted-cp-alerts.yml
mrbojangles3: yes

workloads/whereabouts/metrics-profiles/hosted-cluster-metrics.yml
mrbojangles3: away it goes

workloads/whereabouts/metrics-profiles/hosted-cp-metrics.yml
mrbojangles3: this too, shall leave

workloads/whereabouts/metrics-profiles/mc-metrics.yml
mrbojangles3: also, gone.

workloads/whereabouts/metrics-report.yml
mrbojangles3: I was thinking we would wait till they export something specific.

workloads/whereabouts/run.sh
mrbojangles3: agreed. I added a global exit code for the overlaps, echo it, and then exit with a logical or between the overlaps and the job exit code.

workloads/whereabouts/node-density.yml
afcollins: To me, the answer to this question comes from the results you've observed. If you don't know, then I would omit the threshold here instead.

mrbojangles3: I think having a threshold here has value, cause the multus / whereabouts team can have the joy of seeing a test go from red to green.

workloads/whereabouts/run.sh
jtaleric: well you are modifying the job so we need some way to index the results accordingly. We should think how we can tell our index scripts this was a snowflake/multus test ;)

vishnuchalla: I was just thinking if this can go as a separate canned workload into `kube-burner/kube-burner-ocp` repo and just have the driver script here if planning to run through e2e (i.e for CI)? That way it will be cleaner and shouldn't run into any indexing issues. Also if considering to add this into `kube-burner/kube-burner-ocp` repo, @mrbojangles3 
 can you check if we can reuse the existing `node-density` yaml resources using symbolic links or absolute paths in this workload's config? just a thought 🤔

mrbojangles3:
Looking to check my understanding here. We are thinking it would be best to move the whereabouts testing to be completely inside of kube-burner-ocp, then just calling it as a normal workload of kube-burner inside of e2e. 
So, close this PR, add a new test to kube-burner and then open a newer more focused PR in this repo to enable it. 
Sound right? @jtaleric , @mohit-sheth , @vishnuchalla

vishnuchalla:
> Looking to check my understanding here. We are thinking it would be best to move the whereabouts testing to be completely inside of kube-burner-ocp, then just calling it as a normal workload of kube-burner inside of e2e. So, close this PR, add a new test to kube-burner and then open a newer more focused PR in this repo to enable it. Sound right? @jtaleric , @mohit-sheth , @vishnuchalla 
>> Yes, on this idea.
>
> There is then a separate conversation about landing whereabouts testing in the dedicated OCP version or in the non-ocp version. Whereabouts is a community project.
>> Leaving it upto you to decide if this can go in kube-burner non-ocp or ocp version depending upon the environment that we are planning to run it on.

jtaleric:
> Looking to check my understanding here. We are thinking it would be best to move the whereabouts testing to be completely inside of kube-burner-ocp, then just calling it as a normal workload of kube-burner inside of e2e. So, close this PR, add a new test to kube-burner and then open a newer more focused PR in this repo to enable it. Sound right? @jtaleric , @mohit-sheth , @vishnuchalla
> 
> There is then a separate conversation about landing whereabouts testing in the dedicated OCP version or in the non-ocp version. Whereabouts is a community project.

I think this is being suggested since there is so much overlap/copy-paste from the kube-burner job. 

With the new BYOWL in kube-burner we could remove a bit of complexity introduced here https://github.com/kube-burner/kube-burner-ocp/pull/26

PR #703, opened 2024-06-12
delete aws key if exist

venkataanil:
Delete AWS key if exist and create the new key as we won't be having the pem file for existing key.
Use same name both AWS instance and key.


vishnuchalla (approved):
lgtm

PR #721, opened 2024-09-20
Cleaning up dev ES instance ENV variables

vishnuchalla:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description
Removing dev ES instance ENV vars.
Pre-requisite: https://github.com/openshift/release/pull/57107

- Closes # https://issues.redhat.com/browse/PERFSCALE-3430

## Testing
Through the CI tests

### DIFF
--- a/workloads/api-load/env.sh
+++ b/workloads/api-load/env.sh
@@ -1,7 +1,6 @@
 # Common
 
 TEST_CLEANUP=${TEST_CLEANUP:-true}
-export ES_SERVER=${ES_SERVER:-https://search-perfscale-dev-chmf5l4sh66lvxbnadi4bznl3a.us-west-2.es.amazonaws.com:443}
 export ES_INDEX=ripsaw-api-load
 export ES_SKIP_VERIFY=${ES_SKIP_VERIFY:-true}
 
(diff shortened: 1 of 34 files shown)


workloads/api-load/env.sh
```diff
@@ -1,7 +1,6 @@
 # Common
 
 TEST_CLEANUP=${TEST_CLEANUP:-true}
-export ES_SERVER=${ES_SERVER:-https://search-perfscale-dev-chmf5l4sh66lvxbnadi4bznl3a.us-west-2.es.amazonaws.com:443}
 export ES_INDEX=ripsaw-api-load
```
krishvoor, line 4: Let's have a placeholder with the default value `export ES_SERVER=${ES_SERVER:-""}`

krishvoor (commented):
Minor question.

workloads/api-load/env.sh
vishnuchalla: The code flow and behavior would be the same. Any specific reason to add an empty string?

jtaleric:
Could we say if ES_SERVER == ""; default to local index?

vishnuchalla:
> Could we say if ES_SERVER == ""; default to local index?

IMHO both shouldn't be toggled that way. Local indexing should also happen only when a user opts for it.

dry923 (approved):
LGTM!

PR #727, opened 2024-10-09
[PERFSCALE-3428] Cluster health-check

rsevilla87:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Performing a cluster operator health check before running the actual workload, 
I have considered adding this logic as a step in prow, but I ended up adding this logic here for several reasons like:

- Too many jobs: a new chain will have to be added to every job
- More maintenance: maintaining steps in prow is more difficult than here.

Please leave your thoughts here.

- Related Issue #
- Closes #

### DIFF
--- a/workloads/kube-burner-ocp-wrapper/run.sh
+++ b/workloads/kube-burner-ocp-wrapper/run.sh
@@ -1,6 +1,7 @@
 #!/bin/bash -e
 
 set -e
+oc adm wait-for-stable-cluster --minimum-stable-period=1m --timeout=5m
 source ./egressip.sh
 
 ES_SERVER=${ES_SERVER=https://search-perfscale-dev-chmf5l4sh66lvxbnadi4bznl3a.us-west-2.es.amazonaws.com}
--- a/workloads/network-perf-v2/run.sh
+++ b/workloads/network-perf-v2/run.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -e
+oc adm wait-for-stable-cluster --minimum-stable-period=1m --timeout=5m
 
 source ./env.sh
-source ../../utils/common.sh
 
 # Download k8s-netperf function
 download_netperf() {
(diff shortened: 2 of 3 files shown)


workloads/network-perf-v2/run.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -e
+oc adm wait-for-stable-cluster wait-for-stable-cluster=1m
 
 source ./env.sh
-source ../../utils/common.sh
```
rsevilla87, line 5: Let's consolidate into "echo" rather than "log"

workloads/kube-burner-ocp-wrapper/run.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash -e
 
 set -e
+oc adm wait-for-stable-cluster --minimum-stable-period=1m --timeout=5m
```
vishnuchalla, line 4: Just curious if we need this here, while we already have cluster health [check enabled](https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/ocp.go#L53) by default in ocp-wrapper driver code.

jtaleric, line 4: 5m might be too short. for example, if we want to run kube-burner after an upgrade, I am seeing some of the pods (kubeapi) take > 5 min just to roll out.

rsevilla87: The control plane pods are supposed to be rolled out once the upgrade is finished, another potential issue is the mcp rollout, which runs asynchronously of the `oc adm upgrade` command.

Also, I think we're not running any workload after an upgrade completion

jtaleric: I have tried to set `--minimum-stable-period` to 1m and I saw kube-burner errors due to the cluster still stabilizing, but keeping it at 5min I have had better results, just my current experience...

rsevilla87:
Closing in favor of https://github.com/openshift/release/pull/59579

PR #757, opened 2025-03-12
[PERFSCALE-3773] OKD metadata

josecastillolema:


josecastillolema:
@jtaleric @vishnuchalla @paigerube14 ptal!
@vishnuchalla if/when this lands we will need a new release :grin:

jtaleric:
when I think flavor I think OSP / Instance size... :D Wonder if we should maybe call it stream? or install_stream / install_flavor? that way it is a bit more clear what we are streaming or flavoring ;)

utils/index.sh
```diff
@@ -197,6 +197,15 @@ get_architecture_config(){
     fi
 }
 
+get_stream(){
```
vishnuchalla, line 200: Did you consider using [ADDITIONAL_PARAMS](https://github.com/cloud-bulldozer/e2e-benchmarking/blob/fc43dbff80361e59d812e7c218620f535db6f691/utils/index.sh#L264) for this one from openshift CI itself? Because this is not modifying any existing field and just going about adding one, we need to be mindful about adding more stuff here which is not relevant for all the runs.

vishnuchalla: Good point. The problem with orion filtering you mentioned will anyways exist regardless of adding/updating an existing field. So added it as a planned enhancement to [orion improvements document](https://docs.google.com/document/d/1Cs_GWUKvdinBFze1SmrJHCcqAC_MbEPsjLeMACekuBU/edit?tab=t.0).

And w.r.t stream field, I will let you make that call on its importance for being added as new field in metadata. I would go in the direction as treating it as an additional parameter and having an enhancement in orion to also be able to apply `not match` in the metadata based filtering step.

jtaleric:
We waiting on anything else @josecastillolema

josecastillolema:
> We waiting on anything else @josecastillolema

Nope @jtaleric , no pre-reqs for this one, let's merge as soon as possible, thanks!

josecastillolema:
@jtaleric @vishnuchalla can we consider merging this one? 
Anything missing? :smile:

PR #804, opened 2025-09-24
bump kube-burner-ocp to 1.7.6

mohit-sheth:


afcollins (approved):
Rehearsal passed. Any other jobs others would like run? Can do so from the open PR.

Else we can proceed to merge.

PR #842, opened 2026-04-09
Source egress ip only when workload is used

rsevilla87:
## Type of change

- [x] Optimization

## Description

We shouldn't source egressip.sh unconditionally.


rsevilla87:
ptal @venkataanil

rsevilla87:
ping @venkataanil

mohit-sheth (approved):

